PR #25, opened 2017-05-18
Initial HTTP API support

0xAX:
This commit provides initial support for HTTP API and exposes following API:

  * GET /api/v1/version
  * GET /api/v1/status
  * GET /api/v1/status/accept-new
  * POST /api/v1/status/accept-new/:value


RoadRunnr:
nice work, a few very minor nitpick coming as inline comments

.gitignore
```diff
@@ -10,4 +10,5 @@ rel/example_project
 .rebar
 tetrapak/.local.cache
 *~
-
+_build
+doc
```
RoadRunnr: no newline at end of file

priv/http_api_spec.yaml
```diff
@@ -0,0 +1,72 @@
...
+  StatusAcceptNewRequests:
+    type: "object"
+    properties:
+      acceptNewRequests:
+        type: "boolean"
+  Version:
+    type: "object"
+    properties:
+      version:
+        type: "string"
```
RoadRunnr: no newline at end of file

.gitignore
0xAX: fixed

src/ergw_http_api.erl
```diff
@@ -0,0 +1,67 @@
...
+
+init() ->
+    HttpConfig = application:get_env(ergw, http_api),
+    case HttpConfig of
+        undefined ->
+            lager:debug("HTTP API will not be started because of lack of configuration~n"),
+            ok;
+        {ok, HttpOpts0} ->
+            lager:debug("HTTP API listener options: ~p", [HttpOpts0]),
+            try validate_options(HttpOpts0) of
```
RoadRunnr: i would prefer if this where done as part of the main ergw configuration validation

There should something like in ergw_config.erl:
```
validate_option(http_api, Value) when is_list(Value) ->
    validate_options(fun http_api:validate_options/2, Value);
```

The validate_options needs to be adjusted to follow the scheme used in ergw_config

priv/http_api_spec.yaml
0xAX: this is fixed too

src/ergw_http_api.erl
0xAX: Validation of http config is in the `ergw_config` as you suggested. 

Additionally moved initialization of http listener a little bit further and after config is loaded and initialized. So it should be safe to start http after config is loaded and validated.

src/ergw_http_api.erl
```diff
@@ -0,0 +1,60 @@
...
+            Value
+    end.
+
+validate_options(port, Port) when is_integer(Port) ->
+    Port;
+validate_options(acceptors_num, Acceptors) when is_integer(Acceptors) ->
+    Acceptors;
+validate_options(ip, Ip)
+  when is_tuple(Ip) and tuple_size(Ip) == 4 ->
+    Ip;
```
RoadRunnr: IPv6 only is hardly acceptable for new implementations (only when adding IPv6 support would be a major task, which it shouldn't for cowboy).

Sample: https://github.com/travelping/ergw/blob/master/src/ggsn_gn_proxy.erl#L197

Also, tuple matching is commonly used in ergw instead of is_tuple/tuple_size check for IP addresses.

0xAX: > IPv6 only is hardly acceptable for new implementations (only when adding IPv6 support would be a major task, which it shouldn't for cowboy).

Yes. I didn't mean here to check that we have non IPv6 here. just wanted to check that we are dealing especially with 4-size tuple here. But, yes, will remake this via matching.

priv/http_api_spec.yaml
```diff
@@ -0,0 +1,72 @@
+swagger: '2.0'
+info:
+  title: "ERGW API"
+  version: "1.0.0"
+  description: "ERGW API to interact with ERGW"
```
RoadRunnr: correct capitalization is **erGW**

could you do a quick ```sed -i -e 's/ergw/erGW/gi' priv/http_api_spec.yaml```, please?

src/ergw_http_api.erl
0xAX: updated now

priv/http_api_spec.yaml
0xAX: sure

0xAX: done

priv/http_api_spec.yaml
```diff
@@ -0,0 +1,72 @@
...
+basePath: /api/v1
```
RoadRunnr: And please add:
```
  license:
    name: "GPL 2.0"
    url: "https://www.gnu.org/licenses/#GPL"
```

0xAX: added

src/ergw_http_api.erl
```diff
@@ -0,0 +1,59 @@
...
+        {_, Value} ->
+            Value
+    end.
+
+validate_options(port, Port) when is_integer(Port) ->
+    Port;
+validate_options(acceptors_num, Acceptors) when is_integer(Acceptors) ->
+    Acceptors;
+validate_options(ip, {_, _, _, _} = Ip) ->
+    Ip;
```
RoadRunnr: IPv6 validation is still missing

0xAX: ah, I got you in a wrong way in the previous message. Will do now.

0xAX: I've added `validate_options/2` clause for IPv6. Additionally added `inet` option to listener options (https://github.com/travelping/ergw/pull/25/files#diff-08c7a450c062bff388022d5a3057570eR33) as specified in https://ninenines.eu/docs/en/ranch/1.2/manual/ranch_tcp/ in a case when ipv6 addr is passed

RoadRunnr:
looks good, ready to merge.

I'll pull it into the API branch once I'm done with my outstanding changes.

0xAX:
> I'll pull it into the API branch once I'm done with my outstanding changes.

ok, good 👍

RoadRunnr:
integrated into #26

PR #37, opened 2017-06-21
Serve Prometheus metrics on `/api/v1/metrics`

surik:
Now ergw can provide metrics in Prometheus format on `/api/v1/metrics` endpoint. To configure Prometheus to scape them we should specify`metrics_path`. See example:

    scrape_configs:
        - job_name: 'ergw'
          metrics_path: "api/v1/metrics"
          scrape_interval: 5s
          static_configs:
              - targets: ['localhost:8080']

Depends on ergw configuration `api/v1/metrics` may provide big amount of data. However, it is possible to scrap selective by setting `metrics_path`. Examples:

    metrics_path: "api/v1/metrics/socket"
    metrics_path: "api/v1/metrics/socket/gtp-c"
    metrics_path: "api/v1/metrics/eradius"
    ...

Formatter from exometer to prometheus based on @GalaxyGorilla's [exometer_prometheus](https://github.com/GalaxyGorilla/exometer_prometheus)

PR #38, opened 2017-06-22
Serve Prometheus metrics

RoadRunnr:
This is an alternative version of PR #37.

There are two main problems with that PR:

1. ergw_api should not implement data conversion and certainly nothing front end specific. The code in there is for Erlang internal API's only. Therefore move the encoding specific code into the API handler.
2. exometer:get_values/1 call find_values/1 and then get_value/1 on each one, doing a exometer_get_info/1 for each value again duplicates work that was done before in find_values/1. Instead, use find_values together with get_value and avoid the get_info.

Also, do the re:replace on the metric name only once. Regex are expensive, doing to each component in the name is wasteful.

### DIFF
--- a/METRICS.md
+++ b/METRICS.md
@@ -195,18 +195,18 @@ Counters for the following GTPv2-C Messages types exist:
  * update\_pdn\_connection\_set\_response
  * version\_not\_supported
 
-If the HTTP API has been enable the metrics can be read at `/api/v1/metrics`.
+If the HTTP API has been enable the metrics can be read at `/metrics`.
 Stepping into the result is also possible, e.g.:
 
-    curl -X GET "http://localhost:8080/api/v1/metrics/socket/gtp-c/irx/rx/v1" -H  "accept: application/json"
+    curl -X GET "http://localhost:8080/metrics/socket/gtp-c/irx/rx/v1" -H  "accept: application/json"
 
 Also, erGW can provide metrics in Prometheus format:
 
-    curl -X GET "http://localhost:8080/api/v1/metrics" -H  "accept: text/plain;version=0.0.4"
+    curl -X GET "http://localhost:8080/metrics" -H  "accept: text/plain;version=0.0.4"
 
 or more specific:
 
-    curl -X GET "http://localhost:8080/api/v1/metrics/socket/" -H  "accept: text/plain;version=0.0.4"
+    curl -X GET "http://localhost:8080/metrics/socket/" -H  "accept: text/plain;version=0.0.4"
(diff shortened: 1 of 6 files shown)


src/http_api_handler.erl
```diff
@@ -109,3 +115,99 @@ p2m(Bin) ->
...
+ioize(Something) ->
+    Something.
+
+make_metric_name(Path) ->
+    NameList = lists:join($_, lists:map(fun ioize/1, Path)),
+    NameBin = iolist_to_binary(NameList),
+    re:replace(NameBin, "-|\\.", "_", [global, {return,binary}]).
+
+map_type(undefined)     -> <<"untyped">>;
+map_type(counter)       -> <<"gauge">>;
```
GalaxyGorilla: That does not make sense for your metrics if your counters are monotonically increasing, e.g. you can take "counter" as Prometheus type too if thats the case. In my code I did not force this because exometer counters are not necessarily that way.

METRICS.md
```diff
@@ -199,3 +199,19 @@ If the HTTP API has been enable the metrics can be read at `/api/v1/metrics`.
...
+
+    curl -X GET "http://localhost:8080/api/v1/metrics/socket/" -H  "accept: text/plain;version=0.0.4"
+    # TYPE socket_gtp_c_irx_tx_v2_mbms_session_start_response_retransmit gauge
+    socket_gtp_c_irx_tx_v2_mbms_session_start_response_retransmit 0
+
+    # TYPE socket_gtp_c_irx_tx_v2_mbms_session_start_response_count gauge
+    socket_gtp_c_irx_tx_v2_mbms_session_start_response_count 0
+
+Please read Prometheus [Metric names and labels](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)
+that means all '.' and '-' in metric names which presented above will be replated by '_'.
```
surik: just noticed my typo: replated -> replaced

src/http_api_handler.erl
surik: Makes sense

surik: @GalaxyGorilla Yes, I think we can use counter here

tlnd:
I would rather prefer the endpoint _not_ to be part of `/api/v1`. For sure there should be some stable contract for metrics as well, but I would not make it part of the API version 1 contract.

RoadRunnr:
> I would rather prefer the endpoint not to be part of /api/v1. For sure there should be some stable contract for metrics as well, but I would not make it part of the API version 1 contract.

This PR is not about adding the metrics to the API, they are already in there as JSON. This PR adds the prometheus export format to the API.

Do you have a suggestion for a different path for the metrics API ? Maybe /metrics ?

PR #60, opened 2017-12-04
Fix error log about invalid response in gtp_socket

surik:
The third element in the list of data was missed.
Now we change this to debug level and print message via
gtp_c_lib:fmt_gtp/1.


RoadRunnr:
sigh, yes this should be fixed. But since you want to touch it, we can even do it the right way.

* a duplicate it not an error, so this should be debug
* raw GTP msg prints are hard to read, use gtp_c_lib:fmt_gtp/1 to preprocess them (~p is ok for it, fmt_gtp works like lager:pr)

surik:
@RoadRunnr updated

PR #62, opened 2017-12-06
extend http API with Session API

0xAX:
This patch provides three new http API endpoints:

  * GET /sessions
  * GET /sessions/id/bearer_id
  * DELETE /sessions/id/bearer_id

The first two are to get list of sessions or information about a certain
session by id which can be IMSI or IMEI.

The last one is to terminate a session by the given IMSI or IMEI.

@RoadRunnr The list API just returns all sessions from the ets table, and there is no any manipulations with limit/offset for now

### DIFF
--- a/priv/static/specs/http_api_spec.yaml
+++ b/priv/static/specs/http_api_spec.yaml
@@ -52,6 +52,61 @@ paths:
       responses:
         '200':
           description: returns the metrics
+  /sessions:
+    get:
+      summary: 'get list of sessions'
+      responses:
+        '200':
+          description: 'returns list of the sessions'
+          schema:
+            $ref: '#/definitions/SessionsList'
+  /sessions/{id}/{bearer_id}:
+    get:
+      summary: 'returns information about session with the given IMSI or IMEI'
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+          description: 'session identity (IMSI or IMEI)'
+        - name: bearer_id
+          in: path
+          required: true
+          type: integer
+          default: 5
+          description: 'EPS Bearer Id'
+      responses:
+        '200':
+          description: 'session with the given session identity'
+          schema:
+            $ref: '#/definitions/SessionInfo'
+        '404':
+          description: 'a session with the given session identity is not found'
+          schema:
+            $ref: '#/definitions/SessionItemNotFound'
+    delete:
+      summary: 'stops a session with the given session IMSI or IMEI'
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+          description: 'session identity'
+        - name: bearer_id
+          in: path
+          required: true
+          default: 5
+          type: integer
+          description: 'EPS Bearer Id'
+      responses:
+        '200':
+          description: 'a session with the given session identity is stopped'
+          schema:
+            $ref: '#/definitions/SessionDeleted'
+        '404':
+          description: 'a session with the given session identity is not found'
+          schema:
+            $ref: '#/definitions/SessionNotFound'
 definitions:
   Status:
     type: "object"
@@ -79,3 +134,91 @@ definitions:
     properties:
       version:
         type: "string"
+  SessionsList:
+    type: "object"
+    properties:
+      TotalCount:
+        type: integer
+      Sessions:
+        type: array
+        items:
+          $ref: '#/definitions/SessionInfo'
+  SessionInfo:
+    type: "object"
+    properties:
+      IMSI:
+        type: string
+      IMEI:
+        type: string
+      MSISDN:
+        type: string
+      StartTime:
+        type: string
+        format: date-time
+      LocalGTPEntity:
+        type: object
+        properties:
+          GTP-C:
+            type: object
+            properties:
+              Version:
+                type: string
+              TEID:
+                type: integer
+              IP:
+                type: string
+          GTP-U:
+            type: object
+            properties:
+              Version:
+                type: string
+              TEID:
+                type: integer
+              IP:
+                type: string
+      RemoteGTPEntity:
+        type: object
+        properties:
+          GTP-C:
+            type: object
+            properties:
+              Version:
+                type: string
+              TEID:
+                type: integer
+              IP:
+                type: string
+          GTP-U:
+            type: object
+            properties:
+              Version:
+                type: string
+              TEID:
+                type: integer
+              IP:
+                type: string
+      APN:
+        type: array
+        items:
+          type: string
+  SessionItemNotFound:
+    type: "object"
+    properties:
+      TotalCount:
+        type: integer
+      Session:
+        type: array
+        items:
+          type: string
+  SessionNotFound:
+    type: "object"
+    properties:
+      Result:
+        type: string
+        default: "Session not found"
+  SessionDeleted:
+    type: "object"
+    properties:
+      Result:
+        type: string
+        default: "ok"
--- a/src/http_api_handler.erl
+++ b/src/http_api_handler.erl
@@ -7,9 +7,12 @@
 
 -module(http_api_handler).
 
+-include("include/ergw.hrl").
+-include_lib("stdlib/include/ms_transform.hrl").
+
 -export([init/2, content_types_provided/2,
          handle_request_json/2, handle_request_text/2,
-         allowed_methods/2,
+         allowed_methods/2, delete_resource/2,
          content_types_accepted/2]).
 
 -define(FIELDS_MAPPING, [{accept_new, 'acceptNewRequests'},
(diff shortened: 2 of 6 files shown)


priv/static/specs/http_api_spec.yaml
```diff
@@ -79,3 +134,59 @@ definitions:
...
+        items:
+          $ref: '#/definitions/SessionInfo'
+  SessionInfo:
+    type: "object"
+    properties:
+      IMSI:
+        type: string
+      IMEI:
+        type: string
+      MSIDN:
```
RoadRunnr: should be MSISDN

0xAX: oh, will fix this and in other places too

priv/static/specs/http_api_spec.yaml
```diff
@@ -79,3 +134,59 @@ definitions:
...
+        type: string
+      Version:
+        type: string
+      RemoteControlIp:
```
RoadRunnr: I prefer IP in all captial letters, @mgumz what is the policy in out other API's?

priv/static/specs/http_api_spec.yaml
```diff
@@ -79,3 +134,59 @@ definitions:
...
+        type: string
+      RemoteDataIp:
+        type: string
+      ControlPort:
```
RoadRunnr: this leaks to much internal detail

@mgumz please have a look also.

My idea would be make this seomthing like

```
LocalGTPentity:
  GTP-C:
   Version: {v1,v2}
    TEID: ....
    IP: ...
  GTP-U:
   Version: v1
    TEID: ....
    IP: ...
RemoteGTPentity:
  GTP-C:
   Version: {v1,v2}
    TEID: ....
    IP: ...
  GTP-U:
   Version: v1
    TEID: ....
    IP: ...
```

priv/static/specs/http_api_spec.yaml
```diff
@@ -79,3 +134,59 @@ definitions:
...
+        properties:
+          Name:
+            type: string
+          Ip:
+            type: string
+          Type:
+            type: string
+      StartTime:
```
RoadRunnr: Some more information should also be added:

* QoS parameter
* PDN Type
* RAT Type
* UE IPs
* UE Time Zone
* AMBR
* APN
* User Location Information
* Serving Network

Some of those information have different encodings in GTPv1 and GTPv2

src/http_api_handler.erl
```diff
@@ -93,9 +96,57 @@ handle_request(<<"POST">>, _, json, Req, State) ->
...
+delete_resource(Req, State) ->
+    Path = cowboy_req:path(Req),
+    PathParts = binary:split(Path, <<"/">>, [global]),
+    handle_delete(PathParts, Req, State).
+
+handle_delete([<<>>,<<"api">>,<<"v1">>,<<"sessions">>, Imsi | Rest], Req, State) ->
+    BearerId = case Rest of
+                   [] ->
+                       undefined;
+                   Id ->
```
RoadRunnr:
```
[Id|_] ->
   try binary_to_integer(Id) of
```

would IMHO be nicer

src/http_api_handler.erl
```diff
@@ -93,9 +96,57 @@ handle_request(<<"POST">>, _, json, Req, State) ->
...
+                       try binary_to_integer(lists:nth(1, Id)) of
+                           Num -> Num
+                           catch _:_ -> 5
+                       end
+               end,
+    case find_session(Imsi, BearerId) of
```
RoadRunnr: the rest of ergw uses ```IMSI``` (all capitals)

src/http_api_handler.erl
```diff
@@ -216,3 +267,60 @@ maybe_add_sum(Name, DataPoints, histogram, Payload) ->
...
+
+find_session(Id, BearerId) ->
+    BId = case BearerId of
+              undefined -> 5;
+              _ -> BearerId
+          end,
+    Sockets = application:get_env(ergw, sockets, []),
+    find_session(Sockets, Id, BId).
+
+find_session([], _Id, _BearerId) -> [];
```
RoadRunnr: lookup/get methods in ergw return ```undefined``` or ```{error, not_found}``` in this case. Please use either one of these as well.

src/http_api_handler.erl
```diff
@@ -216,3 +267,60 @@ maybe_add_sum(Name, DataPoints, histogram, Payload) ->
...
+                   end),
+    ets:select(gtp_context_reg, F).
+
+find_session(Id, BearerId) ->
+    BId = case BearerId of
+              undefined -> 5;
+              _ -> BearerId
+          end,
+    Sockets = application:get_env(ergw, sockets, []),
+    find_session(Sockets, Id, BId).
```
RoadRunnr: Is it really necessary to iterate over the sockets? It should be doable to have ets:select without the loop?

src/http_api_handler.erl
```diff
@@ -216,3 +267,60 @@ maybe_add_sum(Name, DataPoints, histogram, Payload) ->
...
+        _ -> find_session(Sockets, SessionIdentity, BearerId)
+    end.
+
+build_session_info(Session) ->
+    SessionInfo = maps:get(context, gtp_context:info(Session)),
+    ControlPort = SessionInfo#context.control_port,
+    DataPort = SessionInfo#context.data_port,
+    #{<<"IMSI">> => SessionInfo#context.imsi,
+      <<"IMEI">> => SessionInfo#context.imei,
+      <<"MSIDN">> => SessionInfo#context.msisdn,
```
RoadRunnr: MSISDN

test/ergw_http_api_SUITE.erl
```diff
@@ -127,6 +142,107 @@ end_per_suite(Config) ->
...
+    TotalCount1 = maps:get(<<"TotalCount">>, Res1),
+    ?equal(1, TotalCount1),
+
+    % Stop session
+    URL2 = get_test_url("/api/v1/sessions/111111111111111/5"),
+    {ok, {_, _, Body2}} = httpc:request(delete, {URL2, [], "", []},
+                                        [], [{body_format, binary}]),
+    Res2 = jsx:decode(Body2, [return_maps]),
+    Result2 = maps:get(<<"Result">>, Res2),
+    ?match(<<"ok">>, Result2),
```
RoadRunnr:
```
?match(#{<<"Result">> := <<"ok">>}, Res2)
```
would IMHO result in more readable errors if the test failes

test/ergw_http_api_SUITE.erl
```diff
@@ -127,6 +142,107 @@ end_per_suite(Config) ->
...
+    % Stop non-existent session
+    URL3 = get_test_url("/api/v1/sessions/111111111111112/5"),
+    {ok, {{_, StatusCode, _}, _, Body3}} = httpc:request(delete, {URL3, [], "", []},
+                                                         [], [{body_format, binary}]),
+    Res3 = jsx:decode(Body3, [return_maps]),
+    Result3 = maps:get(<<"Result">>, Res3),
+    ?match(<<"Session not found">>, Result3),
+    ?match(404, StatusCode),
+
+    % Check that session was destroyed after stop
```
RoadRunnr: you need to wait for the context to be terminated, otherwise this could produce random failures. To make sure all context are gone use ```wait4tunnels(?TIMEOUT)``` (e.g.: https://github.com/travelping/ergw/blob/master/test/pgw_SUITE.erl#L337)

priv/static/specs/http_api_spec.yaml
mgumz: @RoadRunnr i prefer IP as well. it should be treated similar to IMEI, IMSI, GTP etc.

priv/static/specs/http_api_spec.yaml
mgumz: @RoadRunnr LocalGTPEntity ... :)

0xAX:
updated

test/ergw_http_api_SUITE.erl
```diff
@@ -127,6 +148,109 @@ end_per_suite(Config) ->
...
+    Res1 = jsx:decode(Body1, [return_maps]),
+    Session1 = maps:get(<<"Session">>, Res1),
+    ?match(#{<<"MSISDN">> := ?MSISDN, <<"IMSI">> := ?IMSI}, Session1),
+
+    URL2 = get_test_url("/api/v1/sessions/111111111111112"),
+    {ok, {{_, StatusCode, _}, _, Body2}} = httpc:request(get, {URL2, []},
+                                                         [], [{body_format, binary}]),
+    Res2 = jsx:decode(Body2, [return_maps]),
+    ?match(#{<<"Session">> := []}, Res2),
+    ?match(404, StatusCode),
```
RoadRunnr: the StatusCode check could use ```?equal``` and go before the jsx:decode

0xAX: fixed

RoadRunnr:
LGTM, but I would like to have some feedback from @mgumz on the naming LocalGTPEntity and RemoteGTPEntity before merging.

vkatsuba:

vkatsuba:
Сlosed due to the fact that this PR is out of date.

PR #63, opened 2018-01-09
Add redirector mode for GTP socket

surik:
The redirection mode means that a GTP request will be transferred to a GTP backed with saving source IP and port. For the GTP backend, such request looks like it came from particular GTP client directly. 

There are some matching rules to making a decision, see the example of configuration:

```erlang
{rrx, [
    {type,      'gtp-c'},
    {ip,        {192, 168, 1, 100}},
    {reuseaddr, true},
    {redirector, [
        {keep_alive_timeout, 60000},  % (ms) how often redirector socket will send echo_request to nodes to keep connection up
        {retransmit_timeout, 10000},  % (ms) how long request will be memorized to use previously selected node
        {lb_type,            random}, % how to select node from list of nodes. now only `random` is supported
        {nodes, [
            {node, [
                {name, <<"gsn">>},           % name will be user to desribe nodes to make routing in `rules` section
                {keep_alive_version, v1},    % v1 | v2. version to build echo_request
                {address, {192, 168, 1, 16}} % only ipv4 for now
            ]},
            {node, [
                {name, <<"gsn2">>},
                {keep_alive_version, v1},
                {address, {192, 168, 1, 17}}
            ]}
        ]},
        {rules, [
            % both v1 and v2 requests from 192.255.127.127 go to `gsn` decribed above 
            {rule, [
                {conditions, [
                    {sgsn_ip, {192, 255, 127, 127}},
                    {version, [v1, v2]}
                ]},
                {nodes, [<<"gsn">>]}
            ]},
            % v1 requests from 127.127.127.127 with imsi 111111111111111 go randomly to `gsn` or `gsn2`
            {rule, [
                {conditions, [
                    {sgsn_ip, {127, 127, 127, 127}},
                    {imsi,    <<"111111111111111">>},
                    {version, [v1]}
                ]},
                {nodes, [<<"gsn">>, <<"gsn2">>]}
            ]},
            % v2 requests from 127.127.127.127 with imsi 111111111111111 go `gsn2`
            {rule, [
                {conditions, [
                    {sgsn_ip, {127, 127, 127, 127}},
                    {imsi,    <<"111111111111111">>},
                    {version, [v2]}
                ]},
                {nodes, [<<"gsn2">>]}
            ]}
        ]} 
    ]}
]}

```


.travis.yml
```diff
@@ -1,5 +1,5 @@
-sudo: false
-
+dist: trusty
```
RoadRunnr: the  travis changes should go into a separate cset, also make that ```dist: xenial```

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_install:
 script:
   - source test/env.sh
   - ./rebar3 compile
-  - ./rebar3 do xref, ct, coveralls send
+  - sudo env PATH=$PATH ./rebar3 do xref, ct, coveralls send
```
RoadRunnr: ```sudo -E ...``` should do the trick an keep all environment variables

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_install:
 script:
   - source test/env.sh
   - ./rebar3 compile
```
RoadRunnr: traivs has rebar3 support builtin now, so the wget for it and the relative path invocation can be dropped. Simply using ```rebar3 ...`` should work

METRICS.md
```diff
@@ -36,8 +38,9 @@ The path `rtt` is the round trip time histogram for each request/response
 message pair.
 
 The `tx` and `rx` metrics count the number of message of a given type
-transmitted and received. The `timeout` counter exists only for requests
-that require a response.
+transmitted and received. The `rr` metric count the number of message processed
```
RoadRunnr: @mgumz I don't like the ```rr``` short name. Should we use ```lb``` (load balancer) or ```ld``` (load director) instead?

src/gtp_socket.erl
```diff
@@ -162,6 +165,11 @@ validate_option(reuseaddr, Value) when is_boolean(Value) ->
 validate_option(rcvbuf, Value)
   when is_integer(Value) andalso Value > 0 ->
     Value;
+validate_option(redirector, Values) when is_list(Values) ->
+    lager:info("Before ~p~n", [Values]),
```
RoadRunnr: we *all* do forget this kind of debug code ;-), but please remove this instance

test/ergw_http_api_SUITE.erl
```diff
@@ -95,7 +95,7 @@ all() ->
 init_per_suite(Config0) ->
     inets:start(),
     Config1 = [{app_cfg, ?TEST_CONFIG},
-              {handler_under_test, ggsn_gn_proxy}
+              {handlers_under_test, [ggsn_gn_proxy]}
```
RoadRunnr: pure whitespace cleanup changes are ok, but only if you don't mix them with other changes. Make a single cset that collects all white space cleanups if needed.

src/gtp_socket.erl
```diff
@@ -379,6 +399,10 @@ make_request(ArrivalTS, IP, Port,
 family({_,_,_,_}) -> inet;
 family({_,_,_,_,_,_,_,_}) -> inet6.
 
+% this returns family with the version number in the end
+family_v({_,_,_,_}) -> inet4;
+family_v({_,_,_,_,_,_,_,_}) -> inet6.
```
RoadRunnr: I think this is not needed. The only valid use case is currently gen_socket:bind. However, there are enough differences between IPv4 and IPv6 sockets  (and there might be even more) that a separate bind helper makes sense (see gtp_socket:bind_gtp_socket/3).

src/gtp_redirector.erl
```diff
@@ -0,0 +1,334 @@
...
+-include("include/ergw.hrl").
+
+-define(T3, 10 * 1000).
+
+init(IP, #{redirector := #{keep_alive_timeout := KATimeout,
+                           retransmit_timeout := RTTimeout,
+                           lb_type := LBType,
+                           rules := Rules,
+                           nodes := Nodes} = _Opts}) -> 
+    {ok, Socket} = gen_socket:socket(gtp_socket:family(IP), raw, udp),
```
RoadRunnr: this should use the same logic as gtp_socket to create and bind socket and handle socket options.

src/gtp_redirector.erl
```diff
@@ -0,0 +1,334 @@
...
+        {_, {error, no_nodes}} -> 
+            lager:warning("~p: no nodes to redirect request ~p", 
+                          [GtpPort#gtp_port.name, ReqKey]),
+            Redirector;
+        {Cached, {{_, DstIP} = Node, NewRedirector}} ->
+            % if a backend was taken from cache we should not update it in cache this time
+            NewRequests = if Cached == true -> Requests;
+                             true -> ergw_cache:enter(ReqKey, Node, RTTimeout, Requests)
+                          end,
+            Family = gtp_socket:family_v(DstIP),
```
RoadRunnr: this could use gtp_socket:family/1 or a direct case on the type of the DstIP

.travis.yml
surik: I have tried this in the beginning - didn't work. Then I switched to `sudo env PATH=..`. But maybe `dist` is the reason. Will check with `dist: xenial`.

METRICS.md
mgumz: @surik what does the "rr" stand for, actually?

.travis.yml
RoadRunnr: the sudo will also undo the effect of the ```source test/env.sh``` script.

src/gtp_socket.erl
```diff
@@ -45,6 +46,8 @@
 	  requests,
 	  responses,
 
+      redirector  = nil :: gtp_redirector:redirector(),
```
RoadRunnr: pure Erlang code doesn't usually use ```nil```, it uses ```undefined```

METRICS.md
surik: @mgumz  `rr` - redirected requests. It should be something like `rrx` to have the same naming with `rx` and `tx` but because we only have two letters it becomes `rr` :)

.travis.yml
surik: Should I do this in the PR? Maybe we can separate it to no mix redirector with `rebar3`...

surik: I've removed it and switched dist to xenial in #64

METRICS.md
mgumz: why is there a "rr" metric at all? if gtphub runs in redirector mode, all the rx and tx metrics are related only to that mode, right?

surik: Yes, we can use these common `rr` and `tx` after a small fix in metrics code. For now I just remove `rr`.

surik:
I'm going to reopen it for stable/1.x branch

PR #65, opened 2018-01-16
Fix metrics suffix for counters

surik:


test/ggsn_SUITE.erl
```diff
@@ -468,17 +468,27 @@ create_pdp_context_request_resend() ->
...
 
     delete_pdp_context(S, GtpC),
+    ?equal(1, get_value(exometer:get_value(DupId)) - Dup0),
 
     ok = meck:wait(?HUT, terminate, '_', ?TIMEOUT),
     ?match(0, meck:num_calls(?HUT, handle_request, ['_', '_', true, '_'])),
     meck_validate(Config),
     ok.
 
+get_value({ok, DPs}) -> proplists:get_value(value, DPs, -1);
```
RoadRunnr: pgw_SUITE already has a limited exometer check, please use the mechanism for this check and sync exmeter checks in pgw and ggsn SUITE

RoadRunnr: https://github.com/travelping/ergw/blob/master/test/pgw_SUITE.erl#L858

test/ergw_test_lib.erl
```diff
@@ -373,3 +374,17 @@ mkint(C) when $A =< C, C =< $F ->
...
+%%%===================================================================
+%%% Exometer helpers
+%%%===================================================================
+
+match_exo_value(Path, Expect) ->
+    Value = get_exo_value(Path),
+    ?equal(Expect, get_exo_value(Path)).
+
+get_exo_value(Path) ->
+    get_value(exometer:get_value(Path)).
```
RoadRunnr: the get_value wrapper is not needed, in the test cases we want to make sure those exometer value exists.

A simple:
```
get_exo_value(Path) ->
    {ok, Value} = exometer:get_value(Path),
    proplists:get_value(value, Value).
```

is enough.

test/ergw_test_lib.erl
```diff
@@ -373,3 +374,17 @@ mkint(C) when $A =< C, C =< $F ->
     C - $A + 10;
 mkint(C) when $a =< C, C =< $f ->
     C - $a + 10.
+
+%%%===================================================================
+%%% Exometer helpers
+%%%===================================================================
+
+match_exo_value(Path, Expect) ->
+    Value = get_exo_value(Path),
```
RoadRunnr: Value is unused.

PR #66, opened 2018-01-24
Initial support of S11 interface

surik:
It implements SAE-GW as a combination of SGW and PGW with S11 interface.

For now, this is just copy-paste of pgw_s5s8 handler with changed name
and added new interface type.


src/saegw_s11.erl
```diff
@@ -0,0 +1,826 @@
...
+                               } = BearerGroup}
+                          } = IEs} = Request,
+               _Resent,
+               #{context := Context0, aaa_opts := AAAopts, 'Session' := Session} = State) ->
+
+    PAA = maps:get(?'PDN Address Allocation', IEs, undefined),
+
+    FqDataTEID = maps:get({v2_fully_qualified_tunnel_endpoint_identifier, 2}, BearerGroup, undefined),
+    Context1 = case FqDataTEID of
+        #v2_fully_qualified_tunnel_endpoint_identifier{interface_type = ?'S11-C MME'} ->
```
RoadRunnr: Type should be **S1-U eNodeB GTP-U interface**

RoadRunnr:
overall ACK, see two small notes inline.

I'm not sure that dropping support for Change Notification Request, Suspend Notify, and Resume Notify is a good idea.
Missing Suspend/Resume support might cause the connection problems after weak-up.

RoadRunnr:
updated and manually merg

PR #67, opened 2018-02-15
Add DNS helper module to resolve names with NAPTR record

surik:


src/ergw_dns.erl
```diff
@@ -0,0 +1,95 @@
...
+    case application:get_env(ergw, nameservers, []) of
+        [] -> inet_res:resolve(Name, in, naptr);
+        Nameservers ->
+            inet_res:resolve(Name, in, naptr, [{nameservers, Nameservers}])
+    end.
+
+-spec lookup(Name :: string() | inet:ip_address()) -> 
+    list(inet:ip_address()).
+lookup(Name) -> 
+    Response = ?MODULE:naptr(Name),
```
RoadRunnr: is there a reason for the explicit ?MODULE refrence?

surik: to work with `meck` in tests

src/ergw_dns.erl
```diff
@@ -0,0 +1,95 @@
...
+%%====================================================================
+%% API
+%%====================================================================
+
+%% @private
+naptr(Name) ->
+    case application:get_env(ergw, nameservers, []) of
+        [] -> inet_res:resolve(Name, in, naptr);
+        Nameservers ->
+            inet_res:resolve(Name, in, naptr, [{nameservers, Nameservers}])
```
RoadRunnr: @mgumz I would think that we can rely on the DNS config of the host/vm/pod. Do we really need to make the name server configurable?

src/ergw_dns.erl
```diff
@@ -0,0 +1,95 @@
...
+-module(ergw_dns).
+
+-export([naptr/1, lookup/1, lookup/2]).
+
+%%====================================================================
+%% API
+%%====================================================================
+
+%% @private
+naptr(Name) ->
```
RoadRunnr: Sorry, for not noticing earlier.

Erlang DNS will by default use UDP and not retry with TCP if the answer packet gets truncated.
So please, change that to
```
naptr(Name) ->
    NsOpts =
	case setup:get_env(ergw, nameservers) of
	    undefined ->
		[];
	{ok, Nameservers} ->
		[{nameservers, Nameservers}]
	end,
    %% 3GPP DNS answers are almost always large, use TCP by default....
    inet_res:resolve(Name, in, naptr, [{usevc, true} | NsOpts]).
```

surik: Changed

PR #80, opened 2018-05-25
implement GTP load balancer and redirector mode

RoadRunnr:
Based on earlier work by Yury Gargay <yury.gargay@gmail.com>, reworked
to better integrate with the socket logic and reuse the path logic
for keep-alive handling.


vkatsuba:

RoadRunnr:
The type of LB implemented here is standards conform, but does not work in the field because of GTP firewalls.

PR #81, opened 2018-05-28
DNS resolver: fix issue when after NAPTR we get A record with IP

surik:
Despite [RFC 3958](https://tools.ietf.org/html/rfc3958#section-6.4)
says:

```
"A" means that the output of the Rule is a domain name and should
be used to lookup address records for that domain.
```

Result of NAPTR may be A record with IP address as a replacement.
In that case we just need to use that IP as result of lookup procedure.

### DIFF
--- a/src/ergw_dns.erl
+++ b/src/ergw_dns.erl
@@ -60,10 +60,17 @@ match(_, _) -> [].
 %% RFC2915: "A" means that the next lookup should be for either an A, AAAA, or A6 record. 
 %% We only support A for now.
 follow({_Order, _Pref, "a", _Server, _Regexp, Replacement}, Resources) -> 
-    [data(Resource)
-     || Resource <- Resources,
-		    type(Resource) =:= a,
-		    string_equal(domain(Resource), Replacement)];
+    IPs = [data(Resource)
+           || Resource <- Resources,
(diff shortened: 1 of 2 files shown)


mgumz:
what exactly is the issue? the description is a bit short on the problem

RoadRunnr:
From my understanding of RFC 3958 and 3GPP TS 29.303, the replacement field in a NPTR record **can not** be a plain IP address.

[RFC 3958](https://tools.ietf.org/html/rfc3958#section-6.4) clearly states that the output (replacement field) of a record with the "A" flag is a domain name and should be used to look address records for that domain.

surik:
I just updated commit message.

RoadRunnr:
@surik 

> Result of NAPTR may be A record with IP address as a replacement.
> In that case we just need to use that IP as result of lookup procedure.

Where do you get that from?

As I see it, if you get an IP address, then your DNS is misconfigured and needs to be fixed.

surik:
> Where do you get that from?

I'm observing this in the test-lab. I'm not responsible for that DNS.

RoadRunnr:
> I'm observing this in the test-lab. I'm not responsible for that DNS.

Could you provide a sample, please?

PR #90, opened 2018-07-17
Split APN label if there is a dot

aialferov:
A label in a list can come with a dot (or several) in it. This change splits such a label using dot as a separator.


RoadRunnr:
I do agree with coveralls, a test would be nice.

aialferov:
Cheery picked from the stable/1.x, so have tests now.

test/ggsn_SUITE.erl
```diff
@@ -354,6 +356,18 @@ create_pdp_context_request_invalid_apn(Config) ->
     meck_validate(Config),
     ok.
 
+%%--------------------------------------------------------------------
+create_pdp_context_request_dotted_apn() ->
+    [{doc, "Check dotted APN return on Create PDP Context Request"}].
+create_pdp_context_request_dotted_apn(Config) ->
+    S = make_gtp_socket(Config),
+
+    create_pdp_context(dotted_apn, S),
```
surik: I think there is a mistake with cherry-picking it from `stable/1.x` because testing helpers are bit different. The code should look like:

```erlang
create_pdp_context_request_dotted_apn(Config) ->
    create_pdp_context(dotted_apn, Config),
    ...
```

RoadRunnr:
@mgumz merging master into this PR creates a non-linear history.

And why did you do that anyway?

mgumz:
yeah, i rebased in too late hours wrongly.

master now contains both the functional improvement as well as the unit tests.

PR #94, opened 2018-07-31
adjust GC and hibernation for session processes

surik:
This is cherry-picked f1afabe2cbbcd0901a369b9c28637b916d793db8 from `feature/rf` to `stable/1.x`

### DIFF
--- a/src/gtp_dp_kmod.erl
+++ b/src/gtp_dp_kmod.erl
@@ -30,7 +30,9 @@
 %%====================================================================
 
 start_link({Name, SocketOpts}) ->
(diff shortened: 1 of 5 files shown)


src/gtp_dp_kmod.erl
```diff
@@ -29,8 +29,10 @@
 %% API
 %%====================================================================
 
-start_link({Name, SocketOpts}) ->
-    gen_server:start_link(?MODULE, [Name, SocketOpts], []).
+start_link({Name, SocketOpts}) ->
```
RoadRunnr: there is a trailing whitespace in this line

PR #98, opened 2018-09-18
drop the special atom '_' from the permited APN labels

RoadRunnr:
Using a list of APN labels with one of them being the wildcard
label was never supported. The wildcard was always supposed to
be the bare '_' atom (without a list).

The return value from normalize_labels for '_' was broken anyway
because lists:flatmap/2 expects a list as return from the mapping
fun.


RoadRunnr:
@mgumz ping?!

PR #111, opened 2019-04-16
tdf: set reason to normal in close_pdn_context/2 during handling of ASR

0xAX:
in other way usage report information will not be collected in
ergw_gsn_lib:delete_sgi_session/3


0xAX:
Closing PR as the fix was pushed to `feature/tdf` branch already.

PR #115, opened 2019-04-26
Updates for master 2019 04 26

RoadRunnr:


ebengt:
The commit message is 'raise exception in gtp_context request and let the server deal with them'
The server seems to re-raise the exception and I could not see any catch for that. Is this okay?

RoadRunnr:
> The commit message is 'raise exception in gtp_context request and let the server deal with them'
> The server seems to re-raise the exception and I could not see any catch for that. Is this okay?

That is the whole point to not **eat** that exception. The only exception that is supposed to be handled is the `#ctx_err{}`. Other exceptions should crash the server. Before that change, we logged all exceptions and ignored them. That could leaf the gtp context in a broken state.

RoadRunnr:
Travis or Alpine Repo seem to be broken. Docker build failure is unrelated to code changes.

PR #116, opened 2019-05-07
API for deleting multiple GTP contexts

platinumthinker:
In order to drain a PGW from GTP contexts in batches of N, `gtp_context:delete_context/1` needs to be called N times.

This PR implements APIs for this behaviour:

* a public one via the REST-API:  DELETE /api/v1/contexts/<count>
* an ergw-internal one in `ergw_api:delete_contexts/1`

Technical note:  `delete_context` will send a `delete_bearer_request` to the SGW with `reactivation requested`.

### DIFF
--- a/src/ergw_api.erl
+++ b/src/ergw_api.erl
@@ -8,7 +8,7 @@
 -module(ergw_api).
 
 %% API
--export([peer/1, tunnel/1]).
+-export([peer/1, tunnel/1, contexts/1, delete_contexts/1]).
 
 %%%===================================================================
 %%% API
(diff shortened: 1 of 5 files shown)


src/ergw_api.erl
```diff
@@ -51,3 +56,14 @@ collect_contexts(Context, Tunnels) ->
...
     Info = gtp_context:info(Context),
     [Info#{'Process' => Context} | Tunnels].
+
+shuffle(List) ->
+    [Y || {_, Y} <- lists:sort([ {rand:uniform(), El} || El <- List])].
+
+delete_random_peers([], Count) -> {error, Count};
+delete_random_peers(_, 0) -> ok;
+delete_random_peers([{_, Peer} | Tail], Count) when is_pid(Peer) ->
+    ok = gen_server:call(Peer, {unbind, self()}),
```
RoadRunnr: * a `unbind` will not terminate/remove/shutdown a path,
* **it is absolute horrible style and not acceptable to do a direct gen_sever call/cast into another module, never ever do that without having a really, really, really good reason**

src/ergw_api.erl
```diff
@@ -51,3 +56,15 @@ collect_contexts(Context, Tunnels) ->
...
     Info = gtp_context:info(Context),
     [Info#{'Process' => Context} | Tunnels].
+
+shuffle(List) ->
+    [Y || {_, Y} <- lists:sort([ {rand:uniform(), El} || El <- List])].
+
+delete_random_contexts([], Count) -> {error, Count};
+delete_random_contexts(_, 0) -> ok;
+delete_random_contexts([{{_, {Id, _, _}}, Pid} | Tail], Count)
+  when (Id == imsi) or (id == imei), is_pid(Pid) ->
```
RoadRunnr: `id == imei` ??? **This Is Erlang!**

PR #118, opened 2019-05-10
Fix ASR term cause

RoadRunnr:


test/pgw_SUITE.erl
```diff
@@ -1106,6 +1108,36 @@ delete_session_request_resend(Config) ->
     meck_validate(Config),
     ok.
 
+%%--------------------------------------------------------------------
+delete_session_fq_teid() ->
+    [{doc, "Check that a Delete Session Request wirh Sender F-TEID works"}].
```
ebengt: typo. 'wirh' should be 'with'

test/saegw_s11_SUITE.erl
```diff
@@ -298,6 +298,7 @@ common() ->
      create_session_request_x2_handover,
      create_session_request_resend,
      delete_session_request_resend,
+     delete_session_fq_teid,
```
ebengt: How come there is no test case for invalid fq_teid here?

RoadRunnr:
right, new version coming up...

PR #119, opened 2019-05-10
WIP: Add metrics describes ip pools

platinumthinker:
Add 2 metrics:

-  /pool/POOL_TYPE/POOL_FIRST_ADDR/size - size of pool
-  /pool/POOL_TYPE/POOL_FIRST_ADDR/currently_in_use - how many IPs using now

when POOL_TYPE = ipv4/ipv6
         POOL_FIRST_ADDR = like "10_180_0_1" for ipv4 and "8001::" for ipv6

### DIFF
--- a/src/gtp_ip_pool.erl
+++ b/src/gtp_ip_pool.erl
@@ -17,7 +17,7 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
 	 terminate/2, code_change/3]).
 
--record(state, {type, first, last, shift, used, free, used_pool, free_pool}).
+-record(state, {id, type, first, last, shift, used, free, used_pool, free_pool}).
 -record(lease, {ip, client_id}).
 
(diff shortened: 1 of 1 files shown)


src/gtp_ip_pool.erl
```diff
@@ -63,14 +63,20 @@ init(Type, First, Last, Shift) when Last >= First ->
...
 		   first = First,
 		   last = Last,
 		   shift = Shift,
 		   used = 0,
 		   free = Size,
 		   used_pool = UsedTid,
 		   free_pool = FreeTid},
+
+    exo_init_metrics(State),
+    exo_update_pool_size_gauge(State, Size),
```
ebengt: The size is in State, as free. So why have it as argument two?

platinumthinker: because sometimes free != size and it could lead to an errors in feature

PR #132, opened 2019-07-04
bump Erlang to 22.0.5

fholzhauser:


RoadRunnr:
yes, but the tests on travis and gitlab should also be bumped to the same version (where available).

.gitlab-ci. is using the dockerhub repos, so 21.3.8.5 and 22.0.5 should be available.
22.0.5 should also be availabe on travis: https://travis-ci.org/travis-ci/travis-erlang-builder/builds/553654678
21.3.8.5 is not available on travis, stay with 21.3.8.1 there.

PR #138, opened 2019-08-06
stateful context handlers

RoadRunnr:
use stateful (gen_statem) process for handling contexts.


RoadRunnr:
this is ready for review and merge now

PR #144, opened 2019-11-26
switch from lager to Erlang logger

RoadRunnr:


RoadRunnr:
mark this WIP to wait for some of out dependencies to make the switch also

PR #147, opened 2019-12-04
fix ms / µs confusion in metrics

RoadRunnr:
Move recording and conversion into the same place, thus avoiding any
further inconsitencies.


RoadRunnr:
@hwinkel @mgumz @mweinhold-dat I need some feedback and advice on the histogram slots.

We could also make them configurable.

PR #149, opened 2019-12-23
added dot files

vijay-hirani:
Added dot file diagrams


RoadRunnr:
1. many thanks for the documentation
2. please add the rendered output as well (as SVG)
3. if possible, I would prefer plantuml sources instead of graphviz

vijay-hirani:
Thanks for the comments, appreciated. Sure will have a got at plantUMI and produce SVG output diagrams.

RoadRunnr:
> Thanks for the comments, appreciated. Sure will have a got at plantUMI and produce SVG output diagrams.

I had a quick go at the diagrams with plantuml and the diagrams it supports are not really fitting for the supoervisor trees. Leaf them as .dot, but add something to doc/Makefile to generate SVG's from the dot files

PR #152, opened 2020-01-08
Dual Address Bearer Flag support

RoadRunnr:


src/ergw_config.erl
```diff
@@ -30,13 +31,16 @@
...
+-define(ApnDefaults, [{ip_pools, []}, {bearer_type, 'IPv4v6'}, {prefered_bearer_type, 'IPv6'}]).
 -define(DefaultsNodesDefaults, [{vrfs, invalid}, {node_selection, default}]).
 
 -define(is_opts(X), (is_list(X) orelse is_map(X))).
 -define(non_empty_opts(X), ((is_list(X) andalso length(X) /= 0) orelse
 			    (is_map(X) andalso map_size(X) /= 0))).
 
+-define(IS_IPv4(X), (is_tuple(X) andalso tuple_size(X) == 4)).
+-define(IS_IPv6(X), (is_tuple(X) andalso tuple_size(X) == 8)).
+
```
ebengt: Is it necessary to have is_tuple/1 in the guard?
Or is this a way to speed up the test?

RoadRunnr: I'm not sure, it might enough to have the tuple_size/1 guard. The same might apply to the map_size/1 guard.

test/ggsn_SUITE.erl
```diff
@@ -963,25 +978,61 @@ error_indication(Config) ->
...
+		     end),
 
-%%--------------------------------------------------------------------
-ipv4v6_pdp_context_request() ->
-    [{doc, "Check Create PDP Context, Delete PDP Context sequence "
-           "for dual stack IPv4/IPv6 contexts"}].
-ipv4v6_pdp_context_request(Config) ->
-    {GtpC, _, _} = create_pdp_context(ipv4v6, Config),
-    delete_pdp_context(GtpC),
+    {GtpC1, _, _} = create_pdp_context({ipv4, false, default}, Config),
```
ebengt: Could these create_pdp_contetext/2 + delete_pdp_context/1
be simplified with a single create_delete_pdp_context/2 that just did both?
As a way to avoid the GtpC1, GtpC2, ...

RoadRunnr: Other tests do things in between those and I would like to keep the structure somewhat uniform.

PR #159, opened 2020-02-09
Adding UP Inactviity Timer feature

vijay-hirani:
Added Branch to support User Plane Inactivity Timer Feature. Please review.

### DIFF
--- a/src/ggsn_gn.erl
+++ b/src/ggsn_gn.erl
@@ -285,6 +285,13 @@ handle_sx_report(#pfcp{type = session_report_request,
     close_pdp_context(normal, Data),
     {shutdown, Data};
 
+%% User Plane Inactivity Timer expired
+handle_sx_report(#pfcp{type = session_report_request,
+		       ie = #{report_type := #report_type{upir = 1}}},
+		 _State, Data) ->
+    close_pdp_context(normal, Data),
+    {shutdown, Data};
+
 handle_sx_report(#pfcp{type = session_report_request,
 		       ie = #{report_type := #report_type{usar = 1},
 			      usage_report_srr := UsageReport}},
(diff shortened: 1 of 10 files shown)


src/ggsn_gn.erl
```diff
@@ -675,12 +682,19 @@ apply_context_change(NewContext0, OldContext, URRActions,
...
+
 %% 'Idle-Timeout' received from ergw_aaa Session takes precedence over configured one
-add_apn_timeout(Opts, Session, Context) ->
+add_apn_timeout(Opts, Session, Context, PCtx) ->
     SessionWithTimeout = maps:merge(maps:with(['Idle-Timeout'],Opts), Session),
     Timeout = maps:get('Idle-Timeout', SessionWithTimeout),
     ContextWithTimeout = Context#context{'Idle-Timeout' = Timeout},
-    {ContextWithTimeout, SessionWithTimeout}.
+    PCtxTimeout = map_to_up_inactivity_timer(Timeout),
+    PctxWithTimeout = PCtx#pfcp_ctx{up_inactivity_timer = PCtxTimeout},
```
RoadRunnr: updating the pfcp_ctx should not be done in the protocol handlers. ergw_gsn_lib might be a better place to handle it.  The GTP context contains the timeout and is passed to gsn_lib, so all the values should be available.

Also beware that a idle timeout change might happen while the session running (NASREQ RAR can set it), so session modify needs to handle this as well.

src/ggsn_gn.erl
```diff
@@ -675,12 +682,19 @@ apply_context_change(NewContext0, OldContext, URRActions,
     defer_usage_report(URRActions, UsageReport),
     Data#{context => NewContext, pfcp => PCtx}.
 
+map_to_up_inactivity_timer(SessionTimeout) when is_integer(SessionTimeout) ->
+    SessionTimeout div 1000; % UP timer measured in seconds
+map_to_up_inactivity_timer(_SessionTimeout) -> % CP Idle-Timeout infinty = 0 on UP
+    0. % Zero means no UP inactivity timer is activated
```
RoadRunnr: undefined might be better

src/ggsn_gn.erl
vijay-hirani: Move the pfcp_ctx handling to ergw_gsn_lib.

src/ggsn_gn.erl
vijay-hirani: Changed handling of undefined scenario

RoadRunnr:
superseded by #164

PR #160, opened 2020-02-11
add tini PID 1 to ercdf container

0xAX:
Usually, a process spawning subprocesses is responsible for cleaning
them up once they finish. This responsibility propagates down to PID 1.

When finished subprocesses ("childs") are not reaped, they become zombies
which are never vanish. Over time, they can starve the system. Erlang and
Elixir does not take care of spawned children on its own. Thus, any
"remote console" interaction with a running Erlang node will leave behind
a couple of zombies when PID 1 is not taking care of them.
Inside the container, Erlang / Elixir is usually run as PID 1.

This commit adds tini as the new PID 1 of the containers. As a result, all
child processes will be reaped properly.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -33,7 +33,7 @@ FROM alpine:3.10
 WORKDIR /
 RUN     apk update && \
 		apk --no-cache upgrade && \
-		apk --no-cache add zlib ncurses-libs libcrypto1.1 lksctp-tools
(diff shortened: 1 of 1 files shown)


RoadRunnr:
I agree that something like this is needed. But lets have a brief discussion wether  [Tini] https://github.com/krallin/tini), or [S6](https://github.com/just-containers/s6-overlay) is the way to go.

S6 feels more in line with Erlang supervisor behavior, but I haven't investigated it in depth.

PR #183, opened 2020-05-20
Switch to Erlang/OTP 23.0 for docker image

mgumz:
In addition to use Erlang/OTP 23.0 as the runtime for the docker image,
this commit pins the used *rebar3* version to a specific version. This
provides less surprises upon build runs.

By using Erlang/OTP 23.0, the disfunctional "netdev" - config option becomes functional again, due to the upstream bugfix https://github.com/erlang/otp/pull/2600.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,27 +1,28 @@
 # -- build-environment --
 # see https://docs.docker.com/engine/userguide/eng-image/multistage-build/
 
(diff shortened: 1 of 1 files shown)


RoadRunnr:
actually, the .gitlab-ci.yml should have been upgrade to 23.0 as well.

mgumz:
I have applied my changes on top of feature/otp-23 and will close this PR here.

PR #190, opened 2020-06-12
Handle undefined pool in IP allocation

fholzhauser:
The case when there is no pool defined for a type/prefix is expected to return {error, undefined} in ergw_gsn_lib:alloc_ipv4/3 and alloc_ipv6/3.


src/ergw_ip_pool.erl
```diff
@@ -180,16 +180,16 @@ handle_call({get, ClientId, Type, PrefixLen}, _From, #state{pools = Pools} = Sta
     ?LOG(debug, "~w: Allocate IPv: ~p -> ~p, Pool: ~p", [self(), ClientId, Response, Pool]),
     {reply, Response, State#state{pools = maps:put(Key, Pool, Pools)}};
 
+handle_call({get, _ClientId, Type, _PrefixLen}, _From, State)
+  when is_atom(Type) ->
```
ebengt, line 184: Is it worth specifying that Type can be ipv6 or ipv4, not just any atom?

PR #192, opened 2020-06-26
Refactor: move gtp path to gen_statem

vijay-hirani:
Moved the echo timer into gen_statem actions. Have tried to activate and deactivate the echo timer action only where needed

### DIFF
--- a/src/gtp_path.erl
+++ b/src/gtp_path.erl
@@ -250,7 +250,7 @@ handle_event(cast, Msg, _State, _Data) ->
 handle_event(info,{'DOWN', _MonitorRef, process, Pid, _Info}, State0, Data) ->
     State = unregister(Pid, State0, Data),
     case State of
-	#state{echo_timer = 'echo_Stopped'} ->
+	#state{echo_timer = 'stopped'} ->
 	    Actions = echo_timeout_action([], infinity, 'send_echo'),
 	    {next_state, State, Data, Actions};
 	_ ->
--- a/src/gtp_path.erl
+++ b/src/gtp_path.erl
@@ -7,7 +7,7 @@
 
 -module(gtp_path).
 
--behaviour(gen_server).
+-behaviour(gen_statem).
 
 -compile({parse_transform, cut}).
 -compile({no_auto_import,[register/2]}).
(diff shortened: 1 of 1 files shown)


ebengt (approved):
Looks good.

I thought that it was supposed to be two steps. First gen_server => gen_statem. Then adding echo timeout. I could be mistaken.

vijay-hirani:
Looks good.
I thought that it was supposed to be two steps. First gen_server => gen_statem. Then adding echo timeout. I could be mistaken.

VH -> The echo timeout to fire regular 1 minute echos to active nodes is already there. I had added 2 new timers, 1. for monitoring of down nodes and 2. duration to monitor down nodes for the path maintenance.

src/gtp_path.erl
```diff
@@ -168,10 +168,16 @@ handle_event({call, From}, {bind, Pid, RestartCounter}, _State, Data0) ->
 
 handle_event({call, From}, {unbind, Pid}, _State, Data0) ->
     Data = unregister(Pid, Data0),
-    {keep_state, Data, [{reply, From, ok}]};
+    case Data of
+	#{echo_timer := stopped} ->
```
RoadRunnr, line 172: The echo_timer state could IMHO be part of the overall state.

In the conversion to gen_statem, make state a record with a single field `state` (UP/DOWN). And in this cset, add the the echo_timer state into that state.

vijay-hirani: @RoadRunnr 
Do you mean something like: 
-record(state, {state :: {atom(), atom()}}). %  { ['UP' | 'DOWN'], [stopped | echo_to_send | awaiting_response])
Some thing like this will certainly be easier pattern match on the gen_statem state field. 
Can expand the tuple to further add future timers for path maintenance.

RoadRunnr: more like:
```Erlang
record(state, {peer :: 'UP' | 'DOWN',
               echo :: idle | pending)).
```
that only a idea, the real implementation could need different values.

vijay-hirani:
Thanks Bengt, typo on my part. fixed it, running rebar before I update

RoadRunnr:
alternate version merged.

PR #193, opened 2020-07-08
Add Termination-Cause to accounting stop

fholzhauser:


mgumz:
LGTM

RoadRunnr:
easy fix. However, looking at the control flow around this, it seems that the invocation logic for the various session stop events could use some love and cleanup. The TermCause is not the only thing that could use some improvement...

RoadRunnr:
oh, also Unit Test !

PR #199, opened 2020-07-27
rework IP pool to be asynchronous and reusable

RoadRunnr:
* allow IPv4 and IPv6 request to run in parallel
* keep all pool information in the allocated lease
* use stored information when releasing the IP later


RoadRunnr:
reduced this PR to the async pool logic only. Other changes will come in follow up PRs

fholzhauser (approved):

PR #206, opened 2020-08-19
add DHCPv4 pool support

RoadRunnr:
depends on #199 and #205


vkatsuba:

PR #208, opened 2020-08-20
Feature/api for oam

ebengt:
Proof of Concept for 1 configuration detail via HTTP/2 API.


RoadRunnr:
* the handler code does strange things, e.g. not using `malformed_request` to parse the Qs and replicate Qs parsing in strange places
* the management object model seem stage, grouping APNs under `Idle-Timeouts` feels wrong, I would have expected a Management Object per APN.

So, lets forget this PR for the moment and start with an OpenAPI description of the OAM functions and data model.
The API also should adhered the 3GPP principles for service based API (TS 29.500, TS 29.501), for naming convention consult TS 29.501, Sect. 5.1

Note: the 3GPP OAM specification from 3GPP SA5 are not a great help, however if anyone understands them and has ideas -- raise a hand.

ebengt:
Agree with Andreas.

PR #209, opened 2020-09-01
Feature/oam api

ebengt:
Allow change of config item ''apns'' using OAM API.


vkatsuba:

ebengt:
I think that the future for ergw is not in push HTTP API, but in pull. So close.
Last word should come from Andreas.

RoadRunnr:
At least the configs that are implemented in this PR should better be pulled from a NRF or similar.

PR #210, opened 2020-09-10
fix URR processing and change reporting

RoadRunnr:
* fix generation of GTP Change Report IEs
* band-aid fix for not querying URRs when there is event to report


mgumz (commented):
the subject of the commit "don't trigger URR queries when we have on events to report" … is the "on" a typo and actually meant "when we have no events to report"?

RoadRunnr:
> the subject of the commit "don't trigger URR queries when we have on events to report" … is the "on" a typo and actually meant "when we have no events to report"?

fixed

fholzhauser (approved):

PR #212, opened 2020-09-15
Remove unused function

vkatsuba:
Remove `gtp_msg_types/0` function which is not used in the project


mgumz:
@RoadRunnr were the functions in question used as part of testing? or did they serve any other purpose?

ebengt (approved):
If this ever needs to be used, it would be better to have in gtplib.

RoadRunnr (approved):

PR #213, opened 2020-09-17
Configurable TEID-Ranges

vkatsuba:
Implementation of TDD which allows to define the available prefix of Tunnel Endpoint IDs by configuration.
------------------
Implementation of `TEI` calculation was get from [experimental branch](https://github.com/travelping/ergw/blob/exp/stateless/src/ergw_tei_mngr.erl#L177-L188) what was implemented by @RoadRunnr

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,8 @@ This requires a suitable ergw.config, e.g.:
            {[<<"APN1">>], [{vrf, sgi}]}
           ]},
 
+         {teid, {3, 6}}, % {teid, {Prefix, Length}}
+
          {node_selection,
           [{default,
             {static,
--- a/src/ergw_config.erl
+++ b/src/ergw_config.erl
@@ -22,6 +22,7 @@
 	 to_map/1]).
 
 -define(DefaultOptions, [{plmn_id, {<<"001">>, <<"01">>}},
+			 {teid, {0, 0}},
 			 {accept_new, true},
 			 {sockets, []},
 			 {handlers, []},
@@ -254,6 +255,8 @@ validate_option(proxy_map, Opts) ->
     gtp_proxy_ds:validate_options(Opts);
 validate_option(path_management, Opts) when ?is_opts(Opts) ->
     gtp_path:validate_options(Opts);
+validate_option(teid, Value) ->
+    ergw_tei_mngr:validate_option(Value);
 validate_option(Opt, Value)
   when Opt == plmn_id;
        Opt == accept_new;
--- a/src/ergw_tei_mngr.erl
+++ b/src/ergw_tei_mngr.erl
@@ -1,0 +1,148 @@
+%% Copyright 2020, Travelping GmbH <info@travelping.com>
+
+%% This program is free software; you can redistribute it and/or
+%% modify it under the terms of the GNU General Public License
+%% as published by the Free Software Foundation; either version
+%% 2 of the License, or (at your option) any later version.
+
+%% simple implementation of 3GPP TS 29.598 Nudsf service
+%% - functional API according to TS
+%% - not optimized for anything
+
+-module(ergw_tei_mngr).
+
+-behavior(gen_statem).
+
+-compile({parse_transform, cut}).
+
+%% API
+-export([start_link/0, alloc_tei/1, release_tei/2]).
+
+%% gen_statem callbacks
+-export([callback_mode/0, init/1, handle_event/4, terminate/3, code_change/4]).
+
+-export([validate_option/1]).
+
+-include("include/ergw.hrl").
+
+-define(SERVER, ?MODULE).
+-define(SAVE_TIMEOUT, 1000).
+-define(MAX_TRIES, 10).
+-define(POW(X, Y), trunc(math:pow(X, Y))).
+
+-record(data, {node_id, block_id, prefix, prefix_len, tid, save_cnt}).
+
+%%%=========================================================================
+%%%  API
+%%%=========================================================================
+
+start_link() ->
+    gen_statem:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+%% with_keyfun/2
+with_keyfun(#request{gtp_port = Port}, Fun) ->
+    with_keyfun(Port, Fun);
+with_keyfun(#gtp_port{name = Name} = Port, Fun) ->
+    Fun(Name, gtp_context:port_teid_key(Port, _));
+with_keyfun(#pfcp_ctx{name = Name} = PCtx, Fun) ->
+    Fun(Name, ergw_pfcp:ctx_teid_key(PCtx, _)).
+
+%% alloc_tei/1
+alloc_tei(Port) ->
+    with_keyfun(Port, fun alloc_tei/2).
+
+%% alloc_tei/2
+alloc_tei(Name, KeyFun) ->
+    gen_server:call(?SERVER, {alloc_tei, Name, KeyFun}).
+
+%% release_tei/2
+release_tei(Port, TEI) ->
+    with_keyfun(Port, release_tei(_, TEI, _)).
+
+%% release_tei/3
+release_tei(Name, TEI, KeyFun) ->
+    gen_server:call(?SERVER, {release_tei, Name, TEI, KeyFun}).
+
+%%%===================================================================
+%%% Options Validation
+%%%===================================================================
+
+validate_option({Prefix, Len} = Value)
+  when is_integer(Prefix), is_integer(Len),
+       Len >= 0, Len =< 8, Prefix >= 0 ->
+    case ?POW(2, Len) of
+	X when X > Prefix -> Value;
+	_ ->
+	    throw({error, {options, {teid, Value}}})
+    end;
+validate_option(Value) ->
+    throw({error, {options, {teid, Value}}}).
+
+%%%===================================================================
+%%% gen_statem callbacks
+%%%===================================================================
+
+callback_mode() -> handle_event_function.
+
+init([]) ->
+    TID = ets:new(?SERVER, [ordered_set, {keypos, 1}]),
+
+    Data = #data{tid = TID},
+
+    {ok, #{}, Data}.
+
+handle_event({call, From}, {alloc_tei, Name, KeyFun}, State, Data) ->
+    RndState = maybe_init_rnd(Name, State),
+    alloc_tei(From, Name, KeyFun, ?MAX_TRIES, RndState, State, Data);
+
+handle_event({call, From}, {release_tei, _Name, TEI, KeyFun}, _State, #data{tid = TID}) ->
+    ets:delete(TID, KeyFun(TEI)),
+    Actions = [{reply, From, ok}],
+    {keep_state_and_data, Actions};
+
+handle_event({call, From}, Request, _State, _Data) ->
+    Reply = {error, {badarg, Request}},
+    {keep_state_and_data, [{reply, From, Reply}]};
+
+handle_event(_, _Msg, _State, _Data) ->
+    keep_state_and_data.
+
+terminate(_Reason, _State, _Data) ->
+    ok.
+
+code_change(_OldVsn, State, Data, _Extra) ->
+    {ok, State, Data}.
+
+%%%=========================================================================
+%%%  internal functions
+%%%=========================================================================
+
+maybe_init_rnd(Name, State) when is_map_key(Name, State) ->
+    maps:get(Name, State);
+maybe_init_rnd(_, _) ->
+    rand:seed_s(exrop).
+
+alloc_tei_reply(From, Reply, Name, RndState, State, Data) ->
+    Actions = [{reply, From, Reply}],
+    {next_state, State#{Name => RndState}, Data, Actions}.
+
+alloc_tei(From, Name, _KeyFun, 0, RndState, State, Data) ->
+    alloc_tei_reply(From, {error, no_tei}, Name, RndState, State, Data);
+alloc_tei(From, Name, KeyFun, Cnt, RndState0, State, #data{tid = TID} = Data0) ->
+    Data = #data{prefix = Prefix0, prefix_len = PrefixLen} = maybe_update_data(Data0),
+    Prefix = Prefix0 bsl (32 - PrefixLen),
+    Mask = 16#ffffffff bsr PrefixLen,
+    {TEI0, RndState} = rand:uniform_s(16#fffffffe, RndState0),
+    TEI = Prefix + (TEI0 band Mask),
+    case ets:insert_new(TID, {KeyFun(TEI), Name}) of
+	true ->
+	    alloc_tei_reply(From, {ok, TEI}, Name, RndState, State, Data);
+	false ->
+	    alloc_tei(From, Name, KeyFun, Cnt - 1, RndState, State, Data)
+    end.
+
+maybe_update_data(#data{prefix = undefined, prefix_len = undefined} = Data) ->
+    {ok, {Prefix, PrefixLen}} = application:get_env(ergw, teid),
+    Data#data{prefix = Prefix, prefix_len = PrefixLen};
+maybe_update_data(Data) ->
+    Data.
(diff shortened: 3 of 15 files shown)


src/ergw_node_selection.erl
```diff
@@ -248,7 +248,7 @@ lookup_host(Name, dns, NameServers) ->
 			      setelement(2, R, [inet_dns:rr(RR, data) | element(2, R)]);
 			  {in, aaaa} ->
 			      setelement(3, R, [inet_dns:rr(RR, data) | element(3, R)]);
-			  Other ->
+			  _Other ->
```
ebengt, line 251: It might be worth adding _ prefix to unused variables in a separate commit.

.gitignore
```diff
@@ -11,3 +11,6 @@ rel/example_project
 tetrapak/.local.cache
 *~
 _build
+rebar3
```
ebengt, line 14: It might be worth adding these in a separate commit.

README.md
```diff
@@ -105,36 +105,27 @@ DOCKER IMAGES
...
-*The minimum supported Erlang version is 22.3.4.*
-
-Erlang 23.0.3 is the recommended version.
-
-Using rebar:
-
-    # rebar3 compile
-
-RUNNING
+BUILDING & RUNNING
```
ebengt, line 108: Good explanation that could perhaps go into a separate commit.

README.md
```diff
@@ -105,36 +105,27 @@ DOCKER IMAGES
...
 -------
-
+Required:
+* The minimum supported Erlang version is **22.3.4**(*Erlang 23.0.3 is the recommended version.*).
```
jellonek, line 111: Please add space before `(`.

src/gtp_context_reg.erl
```diff
@@ -27,12 +27,15 @@
 %% Include files
 %% --------------------------------------------------------------------
 -include("include/ergw.hrl").
+-include_lib("kernel/include/logger.hrl").
 
 -define(SERVER, ?MODULE).
--record(state, {pids, await_unreg}).
+-record(state, {pids, await_unreg, teid_cfg}).
```
ebengt, line 33: The possible interpretations of cfg (as an acronym) are about 100(?). Writing config makes it easier to understand without being too verbose.

README.md
```diff
@@ -105,36 +105,27 @@ DOCKER IMAGES
...
+```sh
+$ git clone https://github.com/travelping/ergw.git
+$ cd ergw
+$ wget https://s3.amazonaws.com/rebar3/rebar3
+$ chmod u+x ./rebar3
+$ touch ergw.config
 ```
 
-This requires a suitable ergw.config, e.g.:
+Then edit the file that just was created **ergw.config** like described below. This requires a suitable **ergw.config**, e.g.:
```
jellonek, line 128:
```suggestion
then fill just created **ergw.config** file with content like described below providing a suitable configuration, e.g.:
```

src/gtp_context_reg.erl
```diff
@@ -215,14 +219,38 @@ maybe_init_rnd([]) ->
 maybe_init_rnd([{_, RndState}]) ->
     RndState.
 
-alloc_tei(_RndStateKey, _RndState, _KeyFun, 0) ->
+alloc_tei(_RndStateKey, _RndState, _KeyFun, 0, _) ->
     {error, no_tei};
-alloc_tei(RndStateKey, RndState0, KeyFun, Cnt) ->
-    {TEI, RndState} = rand:uniform_s(16#fffffffe, RndState0),
+alloc_tei(RndStateKey, RndState0, KeyFun, Cnt, #{prefix := Prefix0, length := PrefixLen} = Data) ->
```
ebengt, line 224: Data was called TeidCfg in the call. It is easier to understand if it keeps this name.

src/gtp_context_reg.erl
```diff
@@ -215,14 +219,38 @@ maybe_init_rnd([]) ->
...
 	_ ->
-	    alloc_tei(RndStateKey, RndState, KeyFun, Cnt - 1)
+	    alloc_tei(RndStateKey, RndState, KeyFun, Cnt - 1, Data)
     end.
+
+get_teid_cfg() ->
+	get_teid_pref_len(application:get_env(ergw, teid), #{prefix => 0, length => 1}).
+
+get_teid_pref_len({ok, [_|_] = TeidCfg}, DefTEID) ->
+	case proplists:get_value(length, TeidCfg, 0) of
```
ebengt, line 241: It is easier to read code that does not nest case. Perhaps split this into two help functions?

RoadRunnr (changes requested):
This PR could be seen as precursor to https://github.com/travelping/ergw/commit/f977f61a4a46cf287fb9d39c3d9771ef673247ac.
I would therefore recommend to use a stripped down version of that implementation (remove the Nudsf storage).

rebar.config
```diff
@@ -97,7 +97,9 @@
   % test case, but leave it up to the tests to actually create the directories.
   {create_priv_dir, manual_per_tc},
   % Specify that we want the results captured in a Surefire report.
-  {ct_hooks, [{cth_surefire, [{path, "../report.xml"}]}]}
+  {ct_hooks, [{cth_surefire, [{path, "../report.xml"}]}]},
+  % Config for tests
+  {sys_config, ["test/test.config"]}
```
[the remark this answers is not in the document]
vkatsuba: Because I see that in CT we have a lot of places where we use the same values like `{'MS-Primary-DNS-Server', {8,8,8,8}}` and others, this is a simple copy and paste, the firs point why I added is - do set configuration for TEID only once and second point is also if we will be have some immutable configuration we can set configuration only once without putting the same one config in a lot of places in CTs. But if you think that this is bad approach I can remove it.

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,170 @@
...
+get_teid_pref_len({ok, {Prefix, Len}}, _) ->
+	{Prefix, Len};
+get_teid_pref_len({ok, {Len}}, DefaultConfig) when not is_integer(Len); Len < 0; Len > ?MAX_TRIES ->
+	?LOG(warning, "TEID length has wrong configuration: ~p", [Len]),
+	DefaultConfig;
+get_teid_pref_len({ok, {Len}}, _) ->
+	{(fun(X) when X =< 8 -> X; (_) -> 1 end)(?POW(2, Len)), Len};
+get_teid_pref_len(WrongConfig, DefaultConfig) ->
+	?LOG(warning, "TEID has wrong configuration: ~p", [WrongConfig]),
+	DefaultConfig.
```
RoadRunnr, line 170: please look how the other modules implement validate_config methods or implement this in ergw_config.

README.md
vkatsuba: Moved README changes into https://github.com/travelping/ergw/pull/214

README.md
vkatsuba: Moved README changes into https://github.com/travelping/ergw/pull/214

README.md
vkatsuba: Moved README changes into https://github.com/travelping/ergw/pull/214

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,170 @@
...
+	when not is_integer(Len); Len < 0; Len > ?MAX_TRIES; not is_integer(Prefix); Prefix < 0; Prefix > 8 ->
+		?LOG(warning, "TEID length and/or prefix has wrong configuration: ~p", [WrongConfig]),
+		DefaultConfig;
+get_teid_pref_len({ok, {Prefix, Len}}, _) ->
+	{Prefix, Len};
+get_teid_pref_len({ok, {Len}}, DefaultConfig) when not is_integer(Len); Len < 0; Len > ?MAX_TRIES ->
+	?LOG(warning, "TEID length has wrong configuration: ~p", [Len]),
+	DefaultConfig;
+get_teid_pref_len({ok, {Len}}, _) ->
+	{(fun(X) when X =< 8 -> X; (_) -> 1 end)(?POW(2, Len)), Len};
```
vkatsuba, line 167:
```erlang
(fun(X) when X =< 3 -> ?POW(2, X); (_) -> 1 end)(Len).
```

src/ergw_config.erl
```diff
@@ -119,6 +125,21 @@ validate_node_name(Name)
...
+	{Prefix, Len};
+validate_teid({ok, {Len}}, DefaultConfig) when not is_integer(Len); Len < 0; Len > ?MAX_TRIES ->
+	?LOG(warning, "TEID length has wrong configuration: ~p", [Len]),
+	DefaultConfig;
+validate_teid({ok, {Len}}, _) ->
+	{(fun(X) when X =< 3 -> ?POW(2, X); (_) -> 1 end)(Len), Len};
+validate_teid(WrongConfig, DefaultConfig) ->
+	?LOG(warning, "TEID has wrong configuration: ~p", [WrongConfig]),
+	DefaultConfig.
+
```
RoadRunnr, line 142: validations happens before application startup and not later.
Please make this part of the already existing validation runs.

src/ergw_config.erl
```diff
@@ -268,6 +268,8 @@ validate_option(Opt, Value)
        Opt == proxy_map;
        Opt == path_management ->
     throw({error, {options, {Opt, Value}}});
+validate_option(teid, Value) ->
+    ergw_tei_mngr:validate_options({teid, Value});
```
RoadRunnr, line 272: that's not going to work

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,174 @@
...
+
+get_teid_config() ->
+    case application:get_env(ergw, teid) of
+        {ok, {_, _} = Config} ->
+            Config;
+        {ok, {Len}} ->
+            {?POW(2, Len), Len};
+        _ ->
+            {1, 0}
+    end.
```
RoadRunnr, line 174: validation phase should take care of that. The init method should be able to assume that the config key always contains sensible data without this case here.

rebar.config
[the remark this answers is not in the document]
RoadRunnr: Yes. Please, remove that config file.

RoadRunnr (changes requested):
test case for the config validation in config_SUITE is missing.

src/ergw_tei_mngr.erl
vkatsuba: We have a 3 cases:
1 - when set Length and Prefix
2 - when set Length only
3 - when configuration is now a set at all
Here we don't do validation, we try to create default value for Prefix set default Length and Prefix

src/ergw_config.erl
vkatsuba: Can you share - why? It's was tested with some fail config data and looks like validation is working as expected.

RoadRunnr: validate_options expects a list or map, you are passing a tuple

src/ergw_tei_mngr.erl
RoadRunnr: your implementation logic always expect a tuple `{Length, Prefix}`. And that is all it should care about. The logic of having a default or just a single number, should go into the validation step.

src/ergw_config.erl
vkatsuba: `ergw_tei_mngr:validate_options/1` expect tuple, and then try put a list in `ergw_config:validate_options/4` and call looks like `ergw_config:validate_options(fun validate_option/1, [{teid, {Len, Prefix}}], [], list).`. So, into `ergw_config:validate_option/4` is set the list.

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,168 @@
...
+    TID = ets:new(?SERVER, [ordered_set, {keypos, 1}]),
+    NodeId = node(),
+    BlockId = iolist_to_binary(["teid-prefix-", integer_to_list(Prefix)]),
+
+    Data = #data{node_id = NodeId,
+		  block_id = BlockId,
+		  prefix = Prefix,
+		  prefix_len = PrefixLen,
+		  tid = TID,
+		  save_cnt = ?SAVE_COUNTER},
```
RoadRunnr, line 99: node_id, block_id and save_cnt are not really use in this version. Even when we later use them, adding them in this version is not needed.

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,168 @@
...
+
+serialize_data(State, #data{tid = TID}) ->
+    SeedL = maps:fold(fun(K,V,L) -> [{K, rand:export_seed_s(V)}|L] end, [], State),
+    Term =
+	#{seed  => SeedL,
+	  teids => ets:tab2list(TID)},
+    term_to_binary(Term, [{minor_version, 2}, compressed]).
+
+save_data(State, #data{node_id = _NodeId, block_id = _BlockId} = Data) ->
+    _Serialized = serialize_data(State, Data).
```
RoadRunnr, line 149: state_timeout, serialize_data and save_data are not really used in this version.

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,168 @@
...
+
+alloc_tei_reply(From, Reply, Name, RndState, State, Data) ->
+    Actions = [{reply, From, Reply}, state_timeout()],
+    {next_state, State#{Name => RndState}, Data, Actions}.
+
+alloc_tei(From, Name, _KeyFun, 0, RndState, State, Data) ->
+    alloc_tei_reply(From, {error, no_tei}, Name, RndState, State, Data);
+alloc_tei(From, Name, KeyFun, Cnt, RndState0, State,
+	  #data{prefix = Prefix0, prefix_len = PrefixLen, tid = TID} = Data) ->
+    Prefix = Prefix0 bsl (?MAX_TRIES - PrefixLen),
```
RoadRunnr, line 159: ?MAX_TRIES is wrong, that it is 32 is just a coincidence.

src/ergw_tei_mngr.erl
```diff
@@ -0,0 +1,168 @@
...
+-export([callback_mode/0, init/1, handle_event/4, terminate/3, code_change/4]).
+
+-export([validate_option/1]).
+
+-include("include/ergw.hrl").
+
+-define(SERVER, ?MODULE).
+-define(SAVE_COUNTER, 10).
+-define(SAVE_TIMEOUT, 1000).
+-define(MAX_TRIES, 32).
```
RoadRunnr, line 31: 32 is actually quite high for retries, 10 should be by far enough.

test/ggsn_SUITE.erl
```diff
@@ -154,6 +154,8 @@
 		     {ip_pools, ['pool-A']}]}
 		  ]},
 
+		 {teid, {3, 6}},
+
 		 {charging,
```
RoadRunnr, line 159: does the test work without that? the default for teid should have worked (but I didn't test that)

RoadRunnr (changes requested):
there are other things that originate from the Nudsf version of the teid manager that are not needed in this version. They should not be in here.

test/ggsn_SUITE.erl
vkatsuba: Without this setting `{3, 6}` the tests is failed

RoadRunnr: it should work without, the default is {0,0}.
Can you find out why it failes?

vkatsuba: Looks like the value `{0, 0}` is not set by default.

vkatsuba: The default value cannot be set by reason that in `ergw_app:start/2` we call `Pid <- ergw_sup:start_link(),`(where we start `ergw_tei_mngr` - when we call  `ergw_tei_mngr` and try get in the function `init` the configurations, the config is not exist and default value is not set) and then we call `ergw_config:load_config(setup:get_all_env(ergw))` where we are set the default value in configuration. Plus we need also update `ergw:load_config/1` for set default value for new options. Not sure, but looks like the `init` function is not a best plays for getting config of `Prefix, PrefixLen` in current case.

vkatsuba: Fixed.

RoadRunnr (approved):
please squash the changes into a single commit before merging.

PR #215, opened 2020-09-18
Update .gitignore

vkatsuba:
Updated `.gitignore` for ignore few files and folders which can be created and used in development.


mgumz:
LGTM

PR #225, opened 2020-09-29
make sure a added/remove peer does not reset the echo interval

RoadRunnr:
Only down/busy/idle changes should change the echo timeout.

### DIFF
--- a/src/gtp_path.erl
+++ b/src/gtp_path.erl
@@ -226,8 +226,9 @@ init([#gtp_port{name = PortName} = GtpPort, Version, RemoteIP, Args]) ->
 handle_event(enter, #state{peer = Old}, #state{peer = Peer}, Data)
   when Old /= Peer ->
     peer_state_change(Old, Peer, Data),
-    State = peer_state(Peer),
-    {keep_state_and_data, enter_peer_state_action(State, Data)};
+    OldState = peer_state(Old),
+    NewState = peer_state(Peer),
+    {keep_state_and_data, enter_peer_state_action(OldState, NewState, Data)};
 
 handle_event(enter, #state{echo = Old}, #state{peer = Peer, echo = Echo}, Data)
   when Old /= Echo ->
@@ -370,7 +371,9 @@ peer_state_change(_, #peer{state = State}, #{reg_key := RegKey}) ->
 %%% Internal functions
 %%%===================================================================
 
-enter_peer_state_action(State, Data) ->
+enter_peer_state_action(State, State, _Data) ->
+    [];
+enter_peer_state_action(_, State, Data) ->
     [enter_state_timeout_action(State, Data),
      enter_state_echo_action(State, Data)].
 
--- a/test/ggsn_proxy_SUITE.erl
+++ b/test/ggsn_proxy_SUITE.erl
@@ -583,6 +583,7 @@ common() ->
     [invalid_gtp_pdu,
      create_pdp_context_request_missing_ie,
      create_pdp_context_request_accept_new,
+     path_maint,
      path_restart, path_restart_recovery,
      ggsn_broken_recovery,
      path_failure_to_ggsn,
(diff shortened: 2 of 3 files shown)


src/gtp_path.erl
```diff
@@ -370,7 +371,9 @@ peer_state_change(_, #peer{state = State}, #{reg_key := RegKey}) ->
 %%% Internal functions
 %%%===================================================================
 
-enter_peer_state_action(State, Data) ->
+enter_peer_state_action(State, State, _Data) ->
+    [];
+enter_peer_state_action(_, State, Data) ->
```
vkatsuba, line 376: Looks like this clause will create warning like `this clause X cannot match because a previous clause at line N always matches`.

RoadRunnr: It didn't happen when I or the CI build it and I really can't see how this should happen.

test/ggsn_proxy_SUITE.erl
```diff
@@ -711,6 +712,10 @@ init_per_testcase(request_fast_resend, Config) ->
 			     meck:passthrough([Request, Msg, Resent, State, Data])
 		     end),
     Config;
+init_per_testcase(path_maint, Config) ->
+    set_path_timers(#{'echo' => 700}),
```
ebengt, line 716: Is it worth saving the old value in Config? Could be used by end_per_testcase/2 later.

src/gtp_path.erl
vkatsuba: It strange, looks like `rebar` somehow ignores such functions, but such a construction should have created a warning.

src/gtp_path.erl
```diff
@@ -226,8 +226,9 @@ init([#gtp_port{name = PortName} = GtpPort, Version, RemoteIP, Args]) ->
 handle_event(enter, #state{peer = Old}, #state{peer = Peer}, Data)
   when Old /= Peer ->
     peer_state_change(Old, Peer, Data),
-    State = peer_state(Peer),
-    {keep_state_and_data, enter_peer_state_action(State, Data)};
+    OldState = peer_state(Old),
+    NewState = peer_state(Peer),
+    {keep_state_and_data, enter_peer_state_action(OldState, NewState, Data)};
```
vkatsuba, line 231: The `peer_state/1` will return atoms `down/idle/busy` and does nothing else. Maybe make sense clean up the code, remove call `peer_state/1` and also remove function `enter_peer_state_action/3` because looks like we always will return `[]`?

test/ggsn_proxy_SUITE.erl
```diff
@@ -859,6 +867,45 @@ create_pdp_context_request_accept_new(Config) ->
...
+    delete_pdp_context(GtpC0),
+    delete_pdp_context(GtpC1),
+    delete_pdp_context(GtpC2),
+
+    ok = meck:wait(?HUT, terminate, '_', ?TIMEOUT),
+    wait4tunnels(?TIMEOUT),
+
+    meck_validate(Config),
+
+    ?equal(3, map_size(Pings)),
```
ebengt, line 904: Why did proxy send Echo request to 3 different IP? I thought it would only send to GGSN?

test/pgw_proxy_SUITE.erl
```diff
@@ -934,6 +942,45 @@ create_session_overload_response(Config) ->
     ok.
 
 
+%%--------------------------------------------------------------------
+path_maint() ->
+    [{doc, "Check that Echo requests are sent"}].
+path_maint(Config) ->
```
ebengt, line 948: Could it be worth factoring out the similarities between this test and the one in ggsn_proxy_SUITE.erl  ?

ebengt (approved):

test/ggsn_proxy_SUITE.erl
RoadRunnr: Yes, but this whole set_path_timers thing came from an earlier change and I don't want to include too much changes in this simple fix.

src/gtp_path.erl
vkatsuba: You are right, everything is correct. I just poorly looked at the code for pattern matching.

vkatsuba (approved):

test/ggsn_proxy_SUITE.erl
RoadRunnr: The config affects all path that are managed by ergw. And that is PGW->proxy, proxy->client, and proxy->PGW.
The client->proxy path is managed by the test client and that does not have active path maintenance.

test/pgw_proxy_SUITE.erl
RoadRunnr: That would apply to all the protocol handler tests. Yes, it could be done, but it would be a lot of work. Certainly more that can go into this PR

PR #227, opened 2020-09-29
add CODEOWNERS

RoadRunnr:


mgumz (approved):
a bit sparse description about what the effect is … but ok 👍

PR #229, opened 2020-09-30
Implement simple DNS name lookup cache

fholzhauser:


src/ergw_sup.erl
```diff
@@ -30,6 +30,7 @@ start_link() ->
 %% ===================================================================
 
 init([]) ->
+    ok = ergw_node_selection_cache:init(),
```
vkatsuba, line 33: Not sure but maybe make sense create `ergw_node_selection_cache` as worker

fholzhauser: That was the initial implementation indeed. However I wanted to benefit from the ets concurrency for lookup requests without serialization, it ended up being an empty gen_server process just holding the ETS table without any added value.

ebengt (approved):
Good.
For the test you might want to consider adding meck to inet_res:resolve/4, and check that there are only 2 calls.

vkatsuba:
Please, do squash commits in single commit.

vkatsuba (approved):

fholzhauser:
Changed testcase to do more requests and count the resolve meck calls (thanks Bengt for the idea)

PR #231, opened 2020-10-01
Release 2.6.0

mgumz:
This PR prepares the changes required for **ergw**:2.6.0

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,37 @@
 erGW - 3GPP GGSN and PDN-GW in Erlang
 =====================================
 
+Version 2.6.0 - 01 October 2020
+-------------------------------
+
+* feature: provide implementation for `DELETE /api/v1/contexts/:count` for
+  GTP-Proxy setups ("drain")
+* feature: configurable TEID prefix to use
+* feature: async IP assignment for IPv4 and IPv6
+* feature: simple cache for Node Selection results based upon DNS queries
+
+* fix handling DNS responses without IP in Sx node selection
+  (would shutdown of *ergw*)
+* fix Change Reporting Indication/Action (Offline Charging)
+* fix encoding of IE 'Indication'
+* improve GTP Path Maintenance procedures.
+  NOTE: In GTP proxy use of *ergw* the path maintenance now stops forwarding
+        traffic to a non-responsive target PGW after it detects, based on GTP echo
+        probes, that it is not reachable. The PGW path is blocked for a configurable
+        duration (default 1 hour) for forwarding. The configuration of the duration
+        is done with the `down_timeout` parameter of the `path_management` section
+        in the node config (see: `config/ergw-c-node.config` for example)
+
+
+* unify socket definition. 
+  NOTE: this requires changes to the configuration
+  of *ergw*: `sx_socket` is removed from the *ergw* configuration, `sx` socket
+  is now handled in the `sockets` section of *ergw*. See
+  config/ergw-c-node.config for an example.
+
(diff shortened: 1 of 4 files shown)


config/ergw-c-node.config
```diff
@@ -74,7 +57,7 @@
 	   {[<<"apn02">>,<<"example">>,<<"net">>], [{vrf, sgi}]}
 	  ]},
 
-	 {teid, {3, 6}}, % {teid, {Prefix, Length}}
```
mgumz, line 77: @vkatsuba would {3,6} be the default if the config/ergw-c-node.config would be packaged via rebar?

config/ergw-c-node.config
```diff
@@ -107,6 +90,16 @@
 	   }]
 	 },
 
+	 {path_management,
```
mgumz, line 93: @fholzhauser we need a description of how to tune the settings in a gtp-proxy setup for the NEWS.md to describe also the known-issue.

NEWS.md
```diff
@@ -1,6 +1,31 @@
...
+* feature: configurable TEID prefix to use
+* feature: async IP assignment for IPv4 and IPv6
+* feature: simple cache for Node Selection results based upon DNS queries
+
+* fix handling DNS responses without IP in Sx node selection
+  (would shutdown of *ergw*)
+* fix Change Reporting Indication/Action (Offline Charging)
+* fix encoding of IE 'Indication'
+* improve GTP Path Maintenance procedures.
+  NOTE: DESCRIBE KNOWN ISSUE
```
mgumz, line 18: @fholzhauser please describe the known issue and how to mitigate via path-maintenance-config values

config/ergw-c-node.config
RoadRunnr: teh default (when otherwise unconfigured) comes from here https://github.com/travelping/ergw/blob/master/src/ergw_config.erl#L25

ebengt: Default seems to be {0,0}.
In ergw_config.erl

mgumz: yes, but since there "was" a teid-config in config/ergw-c-node.config - that would have been the new default, right?

NEWS.md
```diff
@@ -15,7 +15,13 @@ Version 2.6.0 - 01 October 2020
 * fix Change Reporting Indication/Action (Offline Charging)
 * fix encoding of IE 'Indication'
 * improve GTP Path Maintenance procedures.
-  NOTE: DESCRIBE KNOWN ISSUE
+  NOTE: In GTP hub/proxy use of *ergw* the path maintenance now stops forwarding
```
mgumz, line 18: "hub" does not exist in Github land.

config/ergw-c-node.config
vkatsuba: Yes, this is new default value

ebengt: This is a question of what defaults mean. I think it is the values used when there is nothing configured. In this case {3,6} is configured. So, not default. IMHO.

vkatsuba: > This is a question of what defaults mean. I think it is the values used when there is nothing configured. In this case {3,6} is configured. So, not default. IMHO.

Right. If `{teid, {Prefix, Length}}` is set, will be use value from config file, if not set, then will be `{0, 0}`.

vkatsuba (approved):

fholzhauser (approved):

RoadRunnr (approved):
LGTM

PR #232, opened 2020-10-02
Update *recon* and *observer_cli* to support OTP23

mgumz:
847e514 switched the Erlang runtime to OTP23. Both *recon* and
*observer_cli* support OTP23 in their latest releases (mostly for "alloc
compat").

This commit bumps both dependencies.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -36,8 +36,8 @@
 		       {include_src, false}
 		      ]},
 	       {deps, [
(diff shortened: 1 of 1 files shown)


ebengt (approved):
Good.

Would the suggested:
{deps, [observer_cli]}
(without version number) also be okay?

RoadRunnr (approved):

PR #233, opened 2020-10-05
move all meck handling into init/end_per_testcase

RoadRunnr:
Invoking meck:unload from the test case itself bears the risk of
not executing it when the test case failes.

### DIFF
--- a/test/ggsn_proxy_SUITE.erl
+++ b/test/ggsn_proxy_SUITE.erl
@@ -770,6 +770,14 @@ init_per_testcase(cache_timeout, Config) ->
 init_per_testcase(delete_bearer_requests_multi, Config) ->
     setup_per_testcase(Config),
     Config;
+init_per_testcase(TestCase, Config)
+  when TestCase == proxy_context_selection;
+       TestCase == proxy_context_invalid_selection;
+       TestCase == proxy_context_invalid_mapping;
+       TestCase == proxy_api_v2 ->
+    setup_per_testcase(Config),
+    ok = meck:new(gtp_proxy_ds, [passthrough, no_link]),
+    Config;
 init_per_testcase(_, Config) ->
     setup_per_testcase(Config),
     Config.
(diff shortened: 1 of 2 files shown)


test/ggsn_proxy_SUITE.erl
```diff
@@ -835,6 +843,14 @@ end_per_testcase(delete_bearer_requests_multi, Config) ->
...
     end_per_testcase(Config),
     Config;
+end_per_testcase(TestCase, Config)
+  when TestCase == proxy_context_selection;
+       TestCase == proxy_context_invalid_selection;
+       TestCase == proxy_context_invalid_mapping;
+       TestCase == proxy_api_v2 ->
+    ok = meck:unload(gtp_proxy_ds),
+    end_per_testcase(Config),
+    Config;
```
ebengt, line 853: I think the return value is ignored (apart from special tagged tuples). So just 'ok' is a good return value.

ebengt (approved):

test/pgw_proxy_SUITE.erl
```diff
@@ -895,10 +907,22 @@ end_per_testcase(dns_node_selection, Config) ->
     ok = meck:unload(inet_res),
     end_per_testcase(Config),
     Config;
+end_per_testcase(sx_upf_removal, Config) ->
+    ok = meck:unload(ergw_sx_node),
```
ebengt, line 911: This was not done in previous version.
For symmetry reason it is good that it is added, but might there have been a reason not to do it before?

ebengt (approved):

vkatsuba (approved):

test/ggsn_proxy_SUITE.erl
RoadRunnr: yes, but all end_per_testcase function currently do return config. No need to change that now.

test/pgw_proxy_SUITE.erl
RoadRunnr: the missing unload was actually causing test failures and was the reason I looked at this.

PR #234, opened 2020-10-06
Add radius and diameter example configuration to README

vkatsuba:


README.md
```diff
@@ -53,6 +53,123 @@ See [RADIUS.md](RADIUS.md) for a list of supported Attrbiutes.
...
+            {session, ['Default']},
+            {procedures, [
+                {authenticate, []},
+                {authorize, []},
+                {start, ['RADIUS-Acct']},
+                {interim, ['RADIUS-Acct']},
+                {stop, ['RADIUS-Acct']},
+                {{gx, 'CCR-Initial'},   []},
+                {{gx, 'CCR-Update'},    []},
+                {{gx, 'CCR-Terminate'}, []}
```
RoadRunnr, line 94: It might be worth noting that this does not result in a working setup. We use the Gx API to configure the policies. Either a real Gx server or a fake answer is needed to get working policies.

README.md
```diff
@@ -53,6 +53,123 @@ See [RADIUS.md](RADIUS.md) for a list of supported Attrbiutes.
...
+```
+Example of configuration **epc-ocs** `function` of **DIAMETER**:
+```erlang
+%% ...
+{ergw_aaa, [
+%% ...
+    {functions, [
+        {'epc-ocs', [
+            {handler, ergw_aaa_diameter},
+            {'Origin-Host', <<"CHANGE-ME">>},                           %% <- CHANGE
```
RoadRunnr, line 109: the Origin-Host needs to be resolvable to local IP (either through /etc/hosts or DNS).

As alternative to having a resolvable name, it can also be specified as `{Host :: binary(), IP :: inet:ip_address()}`

README.md
```diff
@@ -53,6 +53,123 @@ See [RADIUS.md](RADIUS.md) for a list of supported Attrbiutes.
...
+    {functions, [
+        {'ergw-pgw-epc-rf', [
+            {handler, ergw_aaa_diameter},
+            {'Origin-Host', <<"CHANGE-ME">>},                           %% <- CHANGE
+            {'Origin-Realm', <<"CHANGE-ME">>},                          %% <- CHANGE
+            {transports, [
+                [
+                    {connect_to, <<"aaa://CHANGE-ME;transport=tcp">>},  %% <- CHANGE
+                    {recbuf,131072},{reuseaddr,false},{sndbuf,131072},  %% <- CHANGE
+                    {unordered,false}                                   %% <- CHANGE
```
RoadRunnr, line 138: unordered is only relevant for SCTP and it is actually prefered that SCTP uses `{unordered, true}` (which is the default).

README.md
mgumz: @RoadRunnr but why is the origin-host entity needs to be resolvable? it is its own "id" … the transport-definition later takes care about the "here is where to connect to" and the entity connected to is able to transport messages to the "id" referenced in here.

i totally get the {connect_to, ...} part of the transport section and that the diameter stack needs to be able to resolve the target. but origin-host?

RoadRunnr: Origin-Host (and Origin-Realm) are use by DIAMETER relays and proxies to route the requests. They are completely permitted to attempt to establish new (not predefined) connections for that. Therefore the name needs to resolvable.

Also, that is how the Erlang DIAMETER stack works. If you don't specify a local IP to bind, it tries to use the Origin-Host for that.

fholzhauser (approved):

PR #237, opened 2020-10-08
Fix broken ergw-c-node.config

vkatsuba:
Remove non valid options and fix syntax issue


ebengt (approved):
Good work finding this.

RoadRunnr (approved):

PR #238, opened 2020-10-10
Add test case for integrity of the configuration file ergw-c-node.config

vkatsuba:


test/config_SUITE.erl
```diff
@@ -1086,3 +1087,15 @@ config(_Config)  ->
     ?error_option(set_cfg_value([path_management, idle_timeout], invalid, ?PGW_PROXY_CONFIG)),
     ?error_option(set_cfg_value([path_management, down_timeout], invalid, ?PGW_PROXY_CONFIG)),
     ok.
+
+ergw_c_node_config() ->
+    [{doc, "Test for integrity of the configuration file ergw-c-node.config"}].
+ergw_c_node_config(_Config) ->
+    {ok, Cwd} = file:get_cwd(),
+    [Path | _] = re:split(Cwd, "_"),
```
ebengt, line 1095: Add a comment to describe what this is doing.
Or change the code to
lists:takewhile( fun(X) -> X =/= "_build" end, filename:split(Path) )

test/config_SUITE.erl
```diff
@@ -1086,3 +1087,15 @@ config(_Config)  ->
...
+    case file:consult(<<Path/binary, "config/ergw-c-node.config">>) of
```
ebengt, line 1096: Consider using filename:join/1 instead of appending binaries.

ebengt (approved):

ebengt (approved):
Looks good.
Could use a comment about where the realative path is actually going.

vkatsuba:
> Looks good.
> Could use a comment about where the realative path is actually going.

This info was added into:
```erlang
ergw_c_node_config() ->
    [{doc, "Test for integrity of the configuration file ergw/config/ergw-c-node.config"}]. %% <-- here
```

ebengt:
Yes, of course. My mistake, I missed it.

test/config_SUITE.erl
```diff
@@ -1086,3 +1087,13 @@ config(_Config)  ->
     ?error_option(set_cfg_value([path_management, idle_timeout], invalid, ?PGW_PROXY_CONFIG)),
     ?error_option(set_cfg_value([path_management, down_timeout], invalid, ?PGW_PROXY_CONFIG)),
     ok.
+
+ergw_c_node_config() ->
+    [{doc, "Test for integrity of the configuration file ergw/config/ergw-c-node.config"}].
+ergw_c_node_config(_Config) ->
+    case file:consult("../../../../config/ergw-c-node.config") of
```
RoadRunnr, line 1094: if we do this (and I'm not yet sure that we should), shouldn't the file also pass a run through `ergw_config:validate_config/1` ?

vkatsuba: I add this test by reason that sometimes we can push incorrect config example, I suppose this test is needed for us. I don't use `ergw_config:validate_config.1` because here https://github.com/travelping/ergw/blob/master/src/ergw_config.erl#L260 we will throw options like `nodes`, `handlers` etc.

RoadRunnr (commented):
I don't like this change, but I would accept it if others (@fholzhauser @ebengt) have valid reasons for including it.

My main concerns:
* ct tests are not supposed to test the integrity of the distribution, they are supposed to test the code only,
* the file path reference in the test only works when executed by rebar3, but ct tests should work no matter what tool is executing them

ebengt:
Perhaps testing that the config example is correct format can be done with .gitlab-ci.yml ? Or .travis.yml ? Instead of CT.

vkatsuba:
Need to come up with a different approach to remove the veto.

PR #239, opened 2020-10-14
Termination cause

vkatsuba:


ebengt (approved):
Looks good.
It would be nice with test of more than 'nornal'.

rebar.config
```diff
@@ -12,7 +12,7 @@
 	{netdata, {git, "https://github.com/RoadRunnr/erl_netdata.git", {ref, "cbd6eaf"}}},
 	{gtplib, {git, "https://github.com/travelping/gtplib.git", {branch, "master"}}},
 	{pfcplib, {git, "https://github.com/travelping/pfcplib.git", {branch, "master"}}},
-	{ergw_aaa, {git, "git://github.com/travelping/ergw_aaa", {tag, "3.5.0"}}},
+	{ergw_aaa, {git, "git://github.com/travelping/ergw_aaa", {branch, "master"}}},
```
vkatsuba, line 15: Changed only for testing.

src/ggsn_gn.erl
```diff
@@ -119,7 +119,7 @@ handle_event({call, From}, terminate_context, _State, Data) ->
 
 handle_event({call, From}, {path_restart, Path}, _State,
 	     #{context := #context{path = Path}} = Data) ->
-    close_pdp_context(normal, Data),
+    close_pdp_context(path_restart, Data),
```
RoadRunnr, line 122: `path_restart` is a bad name for an externally visible the termination cause. This really is a `peer restart`

src/ggsn_gn.erl
```diff
@@ -251,10 +251,10 @@ handle_event(info, _Info, _State, Data) ->
     keep_state_and_data;
 
 handle_event({timeout, context_idle}, stop_session, _state, Data) ->
-    delete_context(undefined, normal, Data);
+    delete_context(undefined, inactivity_timeout, Data);
```
RoadRunnr, line 254: we might distinguish between U and C idle timeout.

src/ggsn_gn.erl
```diff
@@ -279,17 +279,18 @@ handle_pdu(ReqKey, #gtp{ie = PDU} = Msg, _State,
     ergw_gsn_lib:ip_pdu(PDU, Context, PCtx),
     {keep_state, Data}.
 
+%% User Plane Error Indication
 handle_sx_report(#pfcp{type = session_report_request,
 		       ie = #{report_type := #report_type{erir = 1}}},
 	    _State, Data) ->
-    close_pdp_context(normal, Data),
+    close_pdp_context(session_error, Data),
```
RoadRunnr, line 286: wrong naming, A GTP Error Indication, mean the remote peer didn't know about the GTP-U tunnel anymore, could be a remote failure or a peer restart.

RoadRunnr (changes requested):
code looks ok, but the names of the termination causes become externally visible, they needs a high level review and agreement.
And some are plainly wrong (e.g. path_restart)

src/ergw_prometheus.erl
```diff
@@ -105,6 +107,11 @@ declare() ->
 				  {labels, [name, type]},
 				  {buckets, [100, 300, 500, 750, 1000]},
 				  {help, "PFCP Request execution time."}]),
+
+    %% Termination cause metrics
+    prometheus_counter:declare([{name, termination_cause_total},
+                  {labels, [name, type]},
+                  {help, "Total number of termination causes"}]),
```
RoadRunnr, line 114: @mgumz @fholzhauser shouldn't that include version and remote as tags as well? Maybe even direction and message type (for incoming requests)?

src/ggsn_gn.erl
fholzhauser: Yes, that's definitely a good idea. We also still need to improve handling different error scenarios from AAA. I'd propose to address those in a next PR.

src/ergw_prometheus.erl
fholzhauser: Yes, we could consider adding those, however I'm not sure how it'd work consistently with the variety of sources these terminations can be initiated from. Proably it requires further discussion to work it out.

RoadRunnr (approved):

PR #241, opened 2020-10-15
Experiment: add cowboy_swagger

vkatsuba:
Remove 9.1 MB of static files. Covering API as was in http_api_spec.yaml. Provide correct Swagger API for metrics.
---
Reasons o PR:
 * Remove all static files of Swagger
 * Erlang developers can create specs in Erlang maps without any issues
 * Easier to maintain and keep up to date

Link: http://localhost:8000/api-docs/index.html

**P.S.** This is only proposal. This PR can be rejected in any time, coz this is experiment.


src/ergw_http_api_doc.erl
```diff
@@ -0,0 +1,213 @@
...
+        {DefinitionResponseStatus, PropertyResponseStatus},
+        {DefinitionResponseStatusAcceptNew, PropertyResponseStatusAcceptNew}
+    ],
+    
+    %% Add definitions
+    [cowboy_swagger:add_definition(Def, Prop) || {Def, Prop} <- DefinitionsAndProperties],
+
+    % In path
+    InPathStatusAcceptNew = #{
+            name => <<"value">>,
```
vkatsuba, line 77: align code block into left

RoadRunnr:
In general, 3GPP first describes the API using OpenAPI YAML, then we implement this. So, for those APIs there should be no need to add a generated documentation.

For our own APIs, it wouldn't hurt if we thought about them and documented them first in the same way.

I'm not a big fan of auto-documenting what is there (except for very mechanic and repetitive things, like the prometheus metrics).

RoadRunnr:
There are projects that generate functions form OpenAPI specs. Maybe that would be a better way to make sure that we fullfile the promises of specs?

vkatsuba:
> There are projects that generate functions form OpenAPI specs. Maybe that would be a better way to make sure that we fullfile the promises of specs?

Do you mean that - create logic for parse YAML of Swagger specs and try detect if dispatch have the same API as described in YAML and if YAML and API from dispatch will not be equal we will provide an some errors about this?

RoadRunnr:
> > There are projects that generate functions form OpenAPI specs. Maybe that would be a better way to make sure that we fullfile the promises of specs?
> 
> Do you mean that - create logic for parse YAML of Swagger specs and try detect if dispatch have the same API as described in YAML and if YAML and API from dispatch will not be equal we will provide an some errors about this?

more something like this: https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/erlang-server

vkatsuba:
> more something like this: https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/erlang-server
Got it, will be investigated, thanks.

vkatsuba:
Closed by reason that this integration is not needed for current project.

PR #251, opened 2020-11-03
Extend delete contexts REST API

vkatsuba:
- Extend delete contexts REST API for delete all contexts:
  - `/api/v1/contexts/{count}`  - Delete N contexts
  - `/api/v1/contexts` - Delete all contexts
- Fix REST API `/api/v1/contexts/count` 500 error when try delete N contexts when contexts  length is `0`
- Fix CT of `http_api_delete_sessions`

### DIFF
--- a/priv/static/specs/http_api_spec.yaml
+++ b/priv/static/specs/http_api_spec.yaml
@@ -52,6 +52,28 @@ paths:
       responses:
         '200':
           description: returns the metrics
+  /contexts:
+    delete:
+      summary: 'delete all contexts'
+      responses:
+        '200':
+          description: return count of contexts
+          schema:
+            $ref: '#/definitions/DeleteContexts'
+  /contexts/{count}:
+    delete:
+      summary: 'delete N contexts'
+      parameters:
+        - name: count
+          in: path
+          required: true
+          type: integer
+          description: 'number of contexts for deleting'
+      responses:
+        '200':
+          description: return count of contexts
+          schema:
+            $ref: '#/definitions/DeleteContexts'
 definitions:
   Status:
     type: "object"
@@ -62,6 +84,12 @@ definitions:
       plmnId:
         $ref:
           "#/definitions/PLMN"
+  DeleteContexts:
+    type: "object"
+    properties:
+      contexts:
+        type: "integer"
+        default: 0
   PLMN:
     type: "object"
     properties:
--- a/src/http_api_handler.erl
+++ b/src/http_api_handler.erl
@@ -94,14 +94,18 @@ handle_request(<<"POST">>, _, json, Req, State) ->
             {true, Req2, State}
     end;
 
+handle_request(<<"DELETE">>, <<"/api/v1/contexts">>, json, Req, State) ->
+    ok = ergw_api:delete_contexts(all),
+    Response = jsx:encode(#{contexts => length(ergw_api:contexts(all))}),
+    Req2 = cowboy_req:set_resp_body(Response, Req),
+    {true, Req2, State};
+
 handle_request(<<"DELETE">>, <<"/api/v1/contexts/", _/binary>>, json, Req, State) ->
     Value = cowboy_req:binding(count, Req),
     case catch binary_to_integer(Value) of
 	Count when is_integer(Count), Count > 0 ->
 	    ok = ergw_api:delete_contexts(Count),
-	    Contexts = ergw_api:contexts(all),
-
-	    Response = jsx:encode(#{contexts => erlang:length(Contexts)}),
+	    Response = jsx:encode(#{contexts => length(ergw_api:contexts(all))}),
 	    Req2 = cowboy_req:set_resp_body(Response, Req),
 	    {true, Req2, State};
 	_ ->
@@ -109,6 +113,7 @@ handle_request(<<"DELETE">>, <<"/api/v1/contexts/", _/binary>>, json, Req, State
 	    Req3 = cowboy_req:set_resp_body(Response, Req),
 	    {true, Req3, State}
     end;
+
 handle_request(_, _, Req, _, State) ->
     {false, Req, State}.
 
--- a/test/ergw_http_api_SUITE.erl
+++ b/test/ergw_http_api_SUITE.erl
@@ -14,15 +14,8 @@
 -include_lib("gtplib/include/gtp_packet.hrl").
 -include_lib("common_test/include/ct.hrl").
 
--define(CONFIG_UPDATE,
-	[{[sockets, cp, ip], localhost},
-	 {[sockets, irx, ip], test_gsn},
-	 {[sockets, sx, ip], localhost},
-	 {[node_selection, {default, 2}, 2, "topon.gn.ggsn.$ORIGIN"],
-	  {fun node_sel_update/2, final_gsn}},
-	 {[node_selection, {default, 2}, 2, "topon.sx.prox01.$ORIGIN"],
-	  {fun node_sel_update/2, pgw_u01_sx}}
-	]).
+-define(TIMEOUT, 2000).
+-define(HUT, ggsn_gn).
 
 -define(TEST_CONFIG,
 	[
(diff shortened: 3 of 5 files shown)


priv/static/specs/http_api_spec.yaml
```diff
@@ -52,6 +52,28 @@ paths:
...
+      summary: 'delete N contexts'
+      parameters:
+        - name: count
+          in: path
+          required: true
+          type: integer
+          description: 'number of contexts for deleting'
+      responses:
+        '200':
+          description: return count of current contexts after deleting
```
RoadRunnr, line 74: That is not correct. The operation is asynchronous, it only triggers a context shutdown. At the moment the count is take, not all shutdowns need to have completed. The number returned could therefore be higher than expected.

src/http_api_handler.erl
```diff
@@ -89,21 +89,27 @@ handle_request(<<"POST">>, _, json, Req, State) ->
             {true, Req2, State}
     end;
 
+handle_request(<<"DELETE">>, <<"/api/v1/contexts">>, json, Req, State) ->
+    ok = ergw_api:delete_contexts(length(ergw_api:contexts(all))),
```
ebengt, line 93: It could be worth saving 
length(ergw_api:contexts(all))
for later use in this function.

src/http_api_handler.erl
```diff
@@ -89,21 +89,27 @@ handle_request(<<"POST">>, _, json, Req, State) ->
...
     Value = cowboy_req:binding(count, Req),
+    ContextsLen = length(ergw_api:contexts(all)),
     case catch binary_to_integer(Value) of
-	Count when is_integer(Count), Count > 0 ->
+	Count when is_integer(Count), Count > 0, ContextsLen >= Count ->
 	    ok = ergw_api:delete_contexts(Count),
-	    Contexts = ergw_api:contexts(all),
-
-	    Response = jsx:encode(#{contexts => erlang:length(Contexts)}),
+	    Response = jsx:encode(#{contexts => length(ergw_api:contexts(all))}),
```
RoadRunnr, line 104: That mean that the operation will fail if the requested number if higher than the current number of contexts. Why do we need that? Or what is there a problem with requesting to delete 200 context if we have only 100 running?

test/ergw_http_api_SUITE.erl
```diff
@@ -270,6 +271,13 @@ http_api_delete_sessions(Config) ->
     meck_validate(Config),
     ok.
 
+http_api_delete_all_sessions() ->
+    [{doc, "Check DELETE /contexts API"}].
+http_api_delete_all_sessions(_Config) ->
+    Res1 = json_http_request(delete, "/api/v1/contexts"),
+    ?match(#{<<"contexts">> := 0}, Res1),
```
RoadRunnr, line 278: That test is unlikely to be stable. The operation is asynchronous and the deletes are not guaranteed to be complete by the time the API call returns.

RoadRunnr (changes requested):

src/http_api_handler.erl
```diff
@@ -89,21 +89,27 @@ handle_request(<<"POST">>, _, json, Req, State) ->
...
 	    Req2 = cowboy_req:set_resp_body(Response, Req),
 	    {true, Req2, State};
 	_ ->
```
ebengt, line 107: We get to here when Value is not an integer or < 0, and that is fine.
But what about ContextsLen >= Count? Please add a comment why this is not the same as deleting all.

test/ergw_http_api_SUITE.erl
```diff
@@ -270,6 +271,13 @@ http_api_delete_sessions(Config) ->
     meck_validate(Config),
     ok.
 
+http_api_delete_all_sessions() ->
```
ebengt, line 274: Perhaps add tests that delete something?

ebengt (approved):
Looks good

src/http_api_handler.erl
RoadRunnr: No, the values from both calls will no be identical.

RoadRunnr, line 93: It might also be a good idea to have a `ergw_api:delete_contexts(all)` version.

The implementation of delete_contexts uses a random number to choose which context to delete. This is not needed if we really want to remove all contexts.

ebengt: Yes, of course. My mistake.

vkatsuba: Done

src/http_api_handler.erl
```diff
@@ -89,21 +89,30 @@ handle_request(<<"POST">>, _, json, Req, State) ->
...
 	Count when is_integer(Count), Count > 0 ->
 	    ok = ergw_api:delete_contexts(Count),
-	    Contexts = ergw_api:contexts(all),
-
-	    Response = jsx:encode(#{contexts => erlang:length(Contexts)}),
+        Contexts = case length(ergw_api:contexts(all)) - Count of
+            Res when Res > 0 -> Res;
+            _ -> 0
+        end,
+	    Response = jsx:encode(#{contexts => Contexts}),
```
RoadRunnr, line 107: Why? That makes zero sense. DELETE is asynchronous operation and not matter of pretens will change that. Return the correct values and document what they mean in the API description.

test/ergw_http_api_SUITE.erl
```diff
@@ -134,47 +203,265 @@
...
+      fun(_, V) -> set_online_charging(Next, Set, V) end, Cfg);
+set_online_charging([Key|Next], Set, Cfg)
+  when is_map(Cfg) ->
+    Cfg#{Key => set_online_charging(Next, Set, maps:get(Key, Cfg))}.
+
+set_online_charging(Set) ->
+    {ok, Cfg0} = application:get_env(ergw, charging),
+    Cfg = set_online_charging(['_', rulebase, '_'], Set, Cfg0),
+    ok = application:set_env(ergw, charging, Cfg).
+
```
RoadRunnr, line 437: you are adding far too much ballast to to this suite. All this charging stuff and a lot of other things are not needed here.

RoadRunnr (changes requested):
Your latest changes go into the wrong direction.

test/ergw_http_api_SUITE.erl
```diff
@@ -262,11 +394,32 @@ http_api_prometheus_metrics_req(_Config) ->
 http_api_delete_sessions() ->
     [{doc, "Check DELETE /contexts/count API"}].
 http_api_delete_sessions(Config) ->
-    Res1 = json_http_request(delete, "/contexts/0"),
-    ?match(#{<<"contexts">> := 0}, Res1),
+    {GtpC0, _, _} = create_pdp_context(Config),
+    {_, _, _} = create_pdp_context(random, GtpC0),
+    ContextsCount = length(ergw_api:contexts(all)),
```
RoadRunnr, line 399: that is not guaranteed to be `2`

test/ergw_http_api_SUITE.erl
```diff
@@ -262,11 +394,32 @@ http_api_prometheus_metrics_req(_Config) ->
...
+    ?match(2, ContextsCount),
+    Res = json_http_request(delete, "/api/v1/contexts/2"),
+    ct:sleep(80000),
```
RoadRunnr, line 402: 80 seconds is too high, wait4tunnels/1 with a reasonable timeout (the other tests use 2 seconds) should work.

test/ergw_http_api_SUITE.erl
```diff
@@ -262,11 +394,32 @@ http_api_prometheus_metrics_req(_Config) ->
...
 
-    ok = meck:wait(ggsn_gn, terminate, '_', 2000),
-    wait4tunnels(2000),
+http_api_delete_all_sessions() ->
+    [{doc, "Check DELETE /contexts API"}].
+http_api_delete_all_sessions(Config) ->
+    {GtpC0, _, _} = create_pdp_context(Config),
+    {_, _, _} = create_pdp_context(random, GtpC0),
+    ContextsCount = length(ergw_api:contexts(all)),
+    ?match(2, ContextsCount),
```
RoadRunnr, line 416: that is not guaranteed to be `2`

test/ergw_http_api_SUITE.erl
```diff
@@ -262,11 +394,32 @@ http_api_prometheus_metrics_req(_Config) ->
...
+    Res = json_http_request(delete, "/api/v1/contexts"),
+    ct:sleep(80000),
```
RoadRunnr, line 418: timeout ot high as well

RoadRunnr (changes requested):

test/ergw_http_api_SUITE.erl
vkatsuba: at the moment this is the optimal amount of time, there was a check with 60 seconds and looks like contexts is not remove at this time, I tried to find out the reason for such a delay, but unfortunately I could not find the reasons about it... Maybe you can help me understand why we have here such a delay?

RoadRunnr: the API call triggers a delete bearer request from the PGW. If you don't respond to that requests, it will be resent a couple of times. This in the end lead to 60 seconds timeout.
You have to make you actually response, check for a sample check delete_bearer_requests_multi  in the PGW suite and delete_pdp_context_requested in the GGSN suite.

vkatsuba: Thanks a lot. Fixed.

RoadRunnr:
merged (with small edit)

PR #255, opened 2020-11-11
remove reliance on User Plane IP Resource Information IE

RoadRunnr:
When 3GPP Rel. 16 made UPF base TEID assignment mandatory, it also removed
the User Plane IP Resource Information IE. We still relied on that IE
for choosing the IP address type for TEID assignment.

Change that to choose bearer IP types based on the tunnel IP type.

Adjust test cases accordingly.


vkatsuba (approved):

ebengt (approved):
Looks good

fholzhauser (approved):

PR #258, opened 2020-11-13
handle timeouts create session/pdp_context requests

RoadRunnr:
A unresponsive GGSN/PGW would leave the session context in proxy
mode dangling. More over, the proxy TEID would be zero and follow
up context teardown would send out a delete request with TEID = 0.

Handle the timeout and use an state timeout to handle the case where
the first request timed out, but a resent was successful.

Add a CT case.

### DIFF
--- a/src/ggsn_gn_proxy.erl
+++ b/src/ggsn_gn_proxy.erl
@@ -16,7 +16,7 @@
 	 handle_request/5, handle_response/5,
 	 handle_event/4, terminate/3]).
 
--export([delete_context/3, close_context/3]).
+-export([delete_context/4, close_context/4]).
 
 -include_lib("kernel/include/logger.hrl").
 -include_lib("gtplib/include/gtp_packet.hrl").
@@ -177,7 +177,16 @@ handle_event(info, {timeout, _, {delete_pdp_context_request, Direction, _ReqKey,
     {next_state, shutdown, Data};
 
 handle_event(info, _Info, _State, _Data) ->
-    keep_state_and_data.
+    keep_state_and_data;
+
+handle_event(state_timeout, #proxy_request{} = ReqKey, connecting, Data) ->
+    gtp_context:request_finished(ReqKey),
+    delete_forward_session(normal, Data),
+    {next_state, shutdown, Data};
+
+handle_event(state_timeout, _, connecting, Data) ->
+    delete_forward_session(normal, Data),
+    {next_state, shutdown, Data}.
 
 handle_pdu(ReqKey, Msg, _State, Data) ->
     ?LOG(debug, "GTP-U v1 Proxy: ~p, ~p",
@@ -219,11 +228,12 @@ handle_request(_ReqKey, _Request, true, _State, _Data) ->
 
 handle_request(ReqKey,
 	       #gtp{type = create_pdp_context_request,
-		    ie = IEs} = Request, _Resent, _State,
+		    ie = IEs} = Request, _Resent, State,
 	       #{context := Context0, aaa_opts := AAAopts, node_selection := NodeSelect,
 		 left_tunnel := LeftTunnel0,
 		 bearer := #{left := LeftBearer0} = Bearer0,
-		 'Session' := Session} = Data) ->
+		 'Session' := Session} = Data)
+  when State == run; State == connecting ->
     Context = ggsn_gn:update_context_from_gtp_req(Request, Context0),
 
     {LeftTunnel1, LeftBearer1} =
@@ -302,11 +312,13 @@ handle_request(ReqKey,
     forward_request(sgsn2ggsn, ReqKey, Request, RightTunnel,
 		    maps:get(right, Bearer), ProxyContext, Data),
 
-    {keep_state, DataNew};
+    %% 30 second timeout to have enough room for resents
+    Action = [{state_timeout, 30 * 1000, ReqKey}],
+    {next_state, connecting, DataNew, Action};
 
 handle_request(ReqKey,
 	       #gtp{type = update_pdp_context_request} = Request,
-	       _Resent, _State,
+	       _Resent, connected,
 	       #{proxy_context := ProxyContext,
 		 left_tunnel := LeftTunnelOld, right_tunnel := RightTunnelOld,
 		 bearer := #{left := LeftBearerOld, right := RightBearer} = Bearer} = Data)
@@ -336,7 +348,7 @@ handle_request(ReqKey,
 %%
 handle_request(ReqKey,
 	       #gtp{type = update_pdp_context_request} = Request,
-	       _Resent, _State,
+	       _Resent, connected,
 	       #{context := Context,
 		 left_tunnel := LeftTunnel, right_tunnel := RightTunnel,
 		 bearer := #{left := LeftBearer}} = Data)
@@ -349,7 +361,7 @@ handle_request(ReqKey,
 
 handle_request(ReqKey,
 	       #gtp{type = ms_info_change_notification_request} = Request,
-	       _Resent, _State,
+	       _Resent, connected,
 	       #{proxy_context := ProxyContext,
 		 left_tunnel := LeftTunnel, right_tunnel := RightTunnel,
 		 bearer := #{right := RightBearer}} = Data)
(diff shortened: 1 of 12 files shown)


RoadRunnr:
depends on and follows PR #257

src/pgw_s5s8_proxy.erl
```diff
@@ -451,6 +466,15 @@ handle_response(ReqInfo, #gtp{version = v1} = Msg, Request, State, Data) ->
 handle_response(_, _Response, _Request, shutdown, _Data) ->
     keep_state_and_data;
 
+handle_response(#proxy_request{direction = sgw2pgw},
```
vkatsuba, line 469: Looks like this clause and clause at line 473 can be collected as:
```erlang
handle_response(#proxy_request{direction = sgw2pgw}, timeout, #gtp{type = create_session_request}, Req, _) 
  when Req =:= connecting; Req =:= connected ->
    keep_state_and_data;
```

src/pgw_s5s8_proxy.erl
```diff
@@ -621,7 +645,26 @@ handle_response(#proxy_request{request = ReqKey} = _ReqInfo,
     ?LOG(warning, "Unknown Proxy Response ~p", [Response]),
 
     gtp_context:request_finished(ReqKey),
-    keep_state_and_data.
+    keep_state_and_data;
+
+handle_response(_, _, #gtp{type = delete_bearer_request}, shutdown, _) ->
```
vkatsuba, line 650: Looks like can be from line 650 till 667 change to:
```erlang
handle_response(_, _, #gtp{type = Type}, shutdown, _)
  when Type =:= delete_bearer_request; Type =:= delete_session_request ->
    keep_state_and_data;

handle_response({Direction, From}, timeout, #gtp{type = Type}, shutdown_initiated, Data) 
  when Type =:= delete_bearer_request; Type =:= delete_session_request ->
    session_teardown_response({error, timeout}, Direction, From, Data);

handle_response({Direction, From}, #gtp{type = Type}, _, shutdown_initiated, Data) 
  when Type =:= delete_bearer_response; Type =:= delete_session_response ->
    session_teardown_response(ok, Direction, From, Data).
```

src/pgw_s5s8_proxy.erl
```diff
@@ -621,7 +645,26 @@ handle_response(#proxy_request{request = ReqKey} = _ReqInfo,
...
+      keep_state_and_data;
```
vkatsuba, line 651: 6 spaces instead of 4

src/pgw_s5s8_proxy.erl
```diff
@@ -621,7 +645,26 @@ handle_response(#proxy_request{request = ReqKey} = _ReqInfo,
...
+handle_response(_, _, #gtp{type = delete_session_request}, shutdown, _) ->
+      keep_state_and_data;
```
vkatsuba, line 653: 6 spaces instead of 4

src/pgw_s5s8_proxy.erl
```diff
@@ -734,31 +777,48 @@ build_context_request(#tunnel{remote = #fq_teid{teid = TEI}} = Tunnel, Bearer,
...
     RequestIEs = gtp_v2_c:build_recovery(Type, Tunnel, false, RequestIEs0),
-    send_request(Tunnel, ?T3, ?N3, Type, RequestIEs).
+    send_request(Tunnel, ?T3, ?N3, Type, RequestIEs, {Direction, From}),
+    maps:update_with(shutdown, [Direction|_], [Direction], Data);
+initiate_session_teardown(_, _, _, Data) ->
+    Data.
+
+session_teardown_response(Answer, Direction, From, #{shutdown := Shutdown0} = Data) ->
+    case lists:delete(Direction, Shutdown0) of
+	[] ->
```
vkatsuba, line 813: Please, align the code block, it looks like used mix tabs and spaces.

src/ggsn_gn_proxy.erl
```diff
@@ -393,6 +405,15 @@ handle_request(#request{socket = Socket} = ReqKey, Msg, _Resent, _State, _Data)
     gtp_context:request_finished(ReqKey),
     keep_state_and_data.
 
+handle_response(#proxy_request{direction = sgsn2ggsn},
```
vkatsuba, line 408: Looks like this clause and clause at line 412 can be collected as:
```erlang
handle_response(#proxy_request{direction = sgsn2ggsn}, timeout, #gtp{type = create_session_request}, Req, _) 
  when Req =:= connecting; Req =:= connected ->
    keep_state_and_data;

File: test/ggsn_proxy_SUITE.erl [hunk 7]
```diff
@@ -620,8 +621,8 @@ common() ->
      session_accounting,
      sx_upf_reconnect,
      sx_upf_removal,
-     sx_timeout,
-     delete_bearer_requests_multi
+     sx_timeout
+     %% delete_bearer_requests_multi
```
vkatsuba, line 625: Why test case `delete_bearer_requests_multi` is disabled?

vkatsuba:
Not sure but looks like we need to have fix for issue of zero TEID as based on `master` and as based on `2.7.2` for create minor release with fix without global changes of context refactoring, by reason that context refactoring needs to be deep re tested.

src/ggsn_gn_proxy.erl
```diff
@@ -615,33 +643,50 @@ build_context_request(#tunnel{remote = #fq_teid{teid = TEI}} = Tunnel, Bearer,
...
     RequestIEs = gtp_v1_c:build_recovery(Type, Tunnel, false, RequestIEs0),
-    send_request(Tunnel, ?T3, ?N3, Type, RequestIEs).
+    send_request(Tunnel, ?T3, ?N3, Type, RequestIEs, {Direction, From}),
+    maps:update_with(shutdown, [Direction|_], [Direction], Data);
+initiate_pdp_context_teardown(_, _, _, Data) ->
+    Data.
+
+pdp_context_teardown_response(Answer, Direction, From, #{shutdown := Shutdown0} = Data) ->
+    case lists:delete(Direction, Shutdown0) of
+	[] ->
```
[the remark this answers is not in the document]
RoadRunnr: A tab equal 8 spaces, a 12 characters indent therefore takes one tab and 4 spaces. This is intended and how Emacs erlang.el does work.

src/pgw_s5s8_proxy.erl
```diff
@@ -309,11 +319,13 @@ handle_request(ReqKey,
     forward_request(sgw2pgw, ReqKey, Request, RightTunnel,
 		    maps:get(right, Bearer), ProxyContext, DataNew, Data),
 
-    {keep_state, DataNew};
+    %% 30 second timeout to have enough room for resents
+    Action = [{state_timeout, 30 * 1000, ReqKey}],
```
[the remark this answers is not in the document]
RoadRunnr: it is used in exactly 2 places, here and in ggsn_gn_proxy. In both places it is the session establishment code. Unifying these methods is something that I want to follow up in the future (hopefully soonish).
Keeping a bit code duplication till then does not hurt, IMHO.

src/pgw_s5s8_proxy.erl
RoadRunnr: That is left over from handling them differently in an earlier version. I couldn't make my mind up if special handling will be needed or not (or might be needed in the future.
But it doesn't hurt to use the short version for now.

src/pgw_s5s8_proxy.erl
RoadRunnr: Intention was to keep this a identical as possible to the similar code in ggsn_gn_proxy. But using the shorter version doesn't hurt.

src/pgw_s5s8_proxy.erl
RoadRunnr: mix tab+space is indented

test/ggsn_proxy_SUITE.erl
RoadRunnr: that change is actually in PR #257 (which need to be merge first), Reason is explained in the test case.

RoadRunnr:
> Not sure but looks like we need to have fix for issue of zero TEID as based on `master` and as based on `2.7.2` for create minor release with fix without global changes of context refactoring, by reason that context refactoring needs to be deep re tested.

I don't see that. But if a specific commercial project has a requirement for that, it should be raised internally and not discussed here.

vkatsuba (approved):

fholzhauser (approved):

PR #261, opened 2020-11-17
bump OTP to 23.1.2.0 and 22.3.4.12

RoadRunnr:


config/vm.args
```diff
@@ -20,3 +20,6 @@
 
 # Start a pool of asynchronous IO threads
 +A30
+
+# carrier pool config, OTP-16856
++Mucp :
```
vkatsuba, line 25: looks like need add new line here

ebengt (approved):
Looks good

Issue #133, opened 2019-07-09
Broken log output


hsychla:
We see logs with special characters instead of the UE's IPv4 address

```
foo-cnode 06:06:20.139 [error] Sx Node TDF Report on <<6,97,99,99,101,115,115>> for UE IPv4 <<"\n���">> IPv6 undefined
foo-cnode 06:06:20.467 [error] Sx Node TDF Report on <<6,97,99,99,101,115,115>> for UE IPv4 <<"\n���">> IPv6 undefined
foo-cnode 06:06:21.310 [error] Sx Node TDF Report on <<6,97,99,99,101,115,115>> for UE IPv4 <<"\n�~�">> IPv6 undefined
````

vkatsuba:
Fixed.

PR #263, opened 2020-11-17
Fix issue #133 for correct display of IPv4 and IPv6

vkatsuba:
Issue https://github.com/travelping/ergw/issues/133
From the example of task, it can be seen that sometimes, for some reason, the addresses can have binary format like `<<10,65533,126,65533>>` etc. which can break the display of the address and will displayed something like `<<"\ný~ý">>` etc.

### DIFF
--- a/src/ergw_sx_node.erl
+++ b/src/ergw_sx_node.erl
@@ -450,7 +450,8 @@ handle_event({call, From},
 	     _State,
 	     #data{pfcp_ctx = #pfcp_ctx{seid = #seid{dp = SEID}} = PCtx, tdf = Tdf}) ->
     {ok, {tdf, VRF}} = ergw_pfcp:find_urr_by_id(Id, PCtx),
-    ?LOG(debug, "Sx Node TDF Report on ~p for UE IPv4 ~p IPv6 ~p", [VRF, IP4, IP6]),
+    ?LOG(debug, "Sx Node TDF Report on ~p for UE IPv4 ~s IPv6 ~s",
+         [VRF, bin2ntoa(IP4), bin2ntoa(IP6)]),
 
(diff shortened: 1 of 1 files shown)


src/ergw_sx_node.erl
```diff
@@ -450,7 +450,7 @@ handle_event({call, From},
 	     _State,
 	     #data{pfcp_ctx = #pfcp_ctx{seid = #seid{dp = SEID}} = PCtx, tdf = Tdf}) ->
     {ok, {tdf, VRF}} = ergw_pfcp:find_urr_by_id(Id, PCtx),
-    ?LOG(debug, "Sx Node TDF Report on ~p for UE IPv4 ~p IPv6 ~p", [VRF, IP4, IP6]),
+    ?LOG(debug, "Sx Node TDF Report on ~p for UE IPv4 ~p IPv6 ~p", [VRF, ergw_inet:bin2ip(IP4), ergw_inet:bin2ip(IP6)]),
```
RoadRunnr, line 453: If you insist on formating IPs for debug logging, then format to something human readable and not from binary to Erlang tuples.

RoadRunnr (changes requested):

src/ergw_sx_node.erl
vkatsuba: I was think that `{x, x, x, x}` will be normal. What is your preferred format?

RoadRunnr: `inet:ntoa/1`

vkatsuba: Done

fholzhauser (approved):

RoadRunnr (approved):

Issue #245, opened 2020-10-27
walkthrough-saegw.config and walkthrough-saegw.md have outdated configuration format


jellonek:
Mentioned in title files have example configuration with old style sx_socket which was changed during #205

vkatsuba:
Closed. Examples updated.

PR #264, opened 2020-11-17
Fix issue #245: provided correct config example of sockets section

vkatsuba:
Issue https://github.com/travelping/ergw/issues/245.
Providing correct config example for `walkthrough-saegw.config` and for `configuration format`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,8 +297,7 @@ Then fill just created **ergw.config** file with content like described below pr
           ]},
 
          {sockets,
-          [{'cp-socket',
-            [{type, 'gtp-u'},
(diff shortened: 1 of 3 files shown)


ebengt (approved):
Would it not be good to have this (config example) in one file instead of three?
I think markdown has links.

vkatsuba:
> Would it not be good to have this (config example) in one file instead of three?
> I think markdown has links.

Agree, but currently issue in incorrect example of config. Remove few configs or examples from some places this looks like a small refactoring of docs/examples structure. I suppose will be more better if we prepare new PR with this refactoring soon.

jellonek (approved):
LGTM

PR #265, opened 2020-11-18
Remove sx_socket option

vkatsuba:
Remove `sx_socket` option because `sx` socket is now handled in the sockets section of `ergw`


ebengt (approved):
Looks good

PR #267, opened 2020-11-20
Skip CT for master

vkatsuba:
Propose skip CT for master for save time for build images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,12 +46,16 @@ before_script:
     fi
 
 script:
-  - ip -br addr
-  - source test/env.sh
-  - ./rebar3 compile
-  - ./rebar3 do xref, ct
-  - (./rebar3 as test do coveralls send || /bin/true)
-  - (cd _build/test/logs/; zip -1r ../../../ct-logs.zip . )
+  - if [ $TRAVIS_BRANCH != "master" ]; then
+      ip -br addr;
+      source test/env.sh;
+      ./rebar3 compile;
+      ./rebar3 do xref, ct;
+      (./rebar3 as test do coveralls send || /bin/true);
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -75,13 +79,6 @@ jobs:
         - docker push $BUILD_IMAGE:$TAG
         - echo "docker push done"
 
-addons:
-  artifacts:
```
vkatsuba, line 79: If CT skip for master, no make sense upload CT logs cuz artifacts disabled for PR's

RoadRunnr:
I'm not convince that this is a good idea.

Also, WHY? It's not like anyone is impatiently waiting for the CI to pass.

vkatsuba:
> Also, WHY? It's not like anyone is impatiently waiting for the CI to pass.

The main reason that the image can build only after CT pass and this mean if we need to start testing we need to await +-  1 hour, eg if we have 2 fixes we need to await 2 hours for build image from master instead of +- 15 min, second point is, basically we run test twice instead of only one only  for PR.

RoadRunnr:
> > Also, WHY? It's not like anyone is impatiently waiting for the CI to pass.
> 
> The main reason that the image can build only after CT pass and this mean if we need to start testing we need to await +- 1 hour, eg if we have 2 fixes we need to await 2 hours for build image from master instead of +- 15 min, second point is, basically we run test twice instead of only one only for PR.

But travis pushes to docker hub and we should be using quay.io for testing anyhow. I wouldn't mind abandoning docker hub altogether.

Building images for non-tag, master builds with expiry label set could be done before testing is complete. There is no real reason to wait for the tests to finish.
Images with expire label should probably still wait for the test to finish to avoid polluting the registry with broken images.

vkatsuba:
Closed cuz current implementation is not suit.

PR #268, opened 2020-11-24
Fix proxy gw selection when a gw is down

fholzhauser:


vkatsuba:
Good catch! Maybe make sense add some test case for this fix?

fholzhauser:
This is a fairly trivial/safe fix and also needed with some urgency. Extending CT to cover this previously missed scenario is planned in a follow up PR.

vkatsuba (approved):

ebengt (commented):
Looks good

RoadRunnr:
> This is a fairly trivial/safe fix and also needed with some urgency. Extending CT to cover this previously missed scenario is planned in a follow up PR.

The interesting thing is that this line was actually hit by the tests, but the catch all a few lines above hide the problem.

In fact adding a `when is_record(Socket, socket)` to the catch all would have made the unit tests fail.

vkatsuba:
> The interesting thing is that this line was actually hit by the tests, but the catch all a few lines above hide the problem.
> 
> In fact adding a `when is_record(Socket, socket)` to the catch all would have made the unit tests fail.

Currently the checking of socket record already exist in choose_gw/4 however the test cases in `proxy_lib_SUITE` don't crashed by reason that in node_selection section provided one by one node. For raise an error like `function_clause` need add into config section `node_selection` two nodes where first node is invalid and next node is valid, using this configuration example CT in previous version can raise an error like  `function_clause` for case if `choose_gw/4` will be called as `choose_gw(Next, NodeSelect, Socket, Version)` instead of expected `choose_gw(Next, NodeSelect, Version, Socket)`. PR https://github.com/travelping/ergw/pull/270 witch updating of CT was provided.

PR #247, opened 2020-10-29
Improve session termination reason handling WIP

fholzhauser:

PR #271, opened 2020-11-27
Improve session termination reason handling

vkatsuba:
- Must review with PR https://github.com/travelping/ergw_aaa/pull/85 and with https://github.com/travelping/ergw_aaa/pull/95 and https://github.com/travelping/ergw_aaa/pull/99
- Old implementation before refactoring of context https://github.com/travelping/ergw/pull/247


fholzhauser:
I'm not too fond of some of the reason names. I find them too generic (especially from the AAA side). Naming is difficult though, so with their description and having  the handler modules in the metrics I think they are workable. Suggestions for better names would probably be welcome, but If nobody comes up with them, I think it is ok.

test/pgw_SUITE.erl
```diff
@@ -4771,6 +4771,16 @@ gy_async_stop(Config) ->
...
     ok = meck:wait(?HUT, terminate, '_', ?TIMEOUT),
+
+    MfrId = [close_context, peer_reject, ergw_aaa_static],
+    CauseTotal = get_metric(prometheus_counter, termination_cause_total, MfrId, 0),
+    case CauseTotal >= 1 of
+        true ->
+            ct:pal("termination_cause_total [close_context, peer_reject, ergw_aaa_static]: ~p", [CauseTotal]);
+        false ->
+            ct:fail("termination_cause_total error - wrong total: ~p~n", [CauseTotal])
+    end,
```
RoadRunnr, line 4782: why reinventing the wheel? Use the same logic as e.g. in the invalid_gtp_pdu test case.

RoadRunnr (changes requested):
I share @fholzhauser reservations about the naming.

I fell that some errors are still handled far too generic, e.g. a DIAMETER ASR can ocurre on any DIAMETER API. The meaning of an ASR on NASREQ might be different from one on Gy. Same applies for DIAMETER STR.
Anyhow, I'll defer judgement on that to @fholzhauser.

fholzhauser:
I think the functionality covers the use case that motivated it, but adding the origin of the ASR as you suggested at least to the metrics makes perfect sense and does not seem to be too difficult. I'm not sure at this moment that it is necessary to extend the AAA mapping with the origins. We can add that later if needed.

test/pgw_SUITE.erl
vkatsuba: yep... done.

fholzhauser (approved):

test/ggsn_SUITE.erl
```diff
@@ -2089,7 +2089,7 @@ gy_ccr_asr_overlap(Config) ->
     ResponseFun =
 	fun(Request, Result, Avps, SOpts) ->
 		Self ! {'$response', Request, Result, Avps, SOpts} end,
-    AAAReq = #aaa_request{from = ResponseFun, procedure = {gy, 'ASR'},
```
RoadRunnr, line 2092: Whatever triggered that change, that doesn't work.

The procedure is named in the config, changing that to the actual handler is IMHO not correct.

vkatsuba: Done

src/ggsn_gn.erl
```diff
@@ -237,7 +237,7 @@ handle_request(ReqKey,
 handle_request(ReqKey,
 	       #gtp{type = delete_pdp_context_request, ie = _IEs} = Request,
 	       _Resent, _State, #{left_tunnel := LeftTunnel} = Data) ->
-    ergw_gtp_gsn_lib:close_context(normal, Data),
+    ergw_gtp_gsn_lib:close_context({?MODULE, normal}, Data),
```
RoadRunnr, line 240: I'm not too happy about this. Passing the ?MODULE like this, means it end up in the `api` label of the prometheus metric and I really don't see how that would be ok. Also, using 'gn' as API also doesn't really seem sensible.

@fholzhauser any ideas?

RoadRunnr (approved):
LGTM

PR #272, opened 2020-12-01
fix reading from socket error queue

RoadRunnr:
A select info is generated whenever there is something to read from
the socket, in C terms `(POLLIN | POLLERR)`. There is no distinction
whether that is normal read or errqueue read (POLLERR).
This adds a heuristic to read the errqueue should we get no data from
a normal socket read.

### DIFF
--- a/src/ergw_gtp_c_socket.erl
+++ b/src/ergw_gtp_c_socket.erl
@@ -331,23 +331,29 @@ make_request(ArrivalTS, IP, Port, Msg, #state{gtp_socket = Socket, info = Info})
     ergw_gtp_socket:make_request(ArrivalTS, IP, Port, Msg, Socket, Info).
 
 handle_input(Socket, Info, #state{burst_size = BurstSize} = State) ->
-    handle_input(Socket, Info, BurstSize, State).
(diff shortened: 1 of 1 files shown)


src/ergw_gtp_c_socket.erl
```diff
@@ -406,13 +412,15 @@ handle_socket_error(Error, IP, _Port, _State) ->
...
+    case socket:recvmsg(Socket, [errqueue], 0) of
 	{ok, #{addr := #{addr := IP, port := Port}, ctrl := Ctrl}} ->
 	    lists:foreach(handle_socket_error(_, IP, Port, State), Ctrl),
 	    ok;
 
-	{select, SelectInfo} ->
-	    socket:cancel(Socket, SelectInfo);
+	{error, timeout} ->
+	    ok;
+	{error, ewouldblock} ->
```
vkatsuba, line 422: Do we need log these errors?

RoadRunnr: no, they are expected and actually mean the same.

vkatsuba (approved):

PR #273, opened 2020-12-03
use random udp source port for GTP-C requests

RoadRunnr:
Both GTP v1 and v2 have similar definitions for UDP source port
of GTP requests:

> The UDP Source Port is a locally allocated port number at the sending GSN/RNC.

Doing so has advantages when the receiver is using RSSI, NAT gateways
need to map those requests and are helpful for outgoing proxies.


RoadRunnr:
This PR includes and depends on PR #272

vkatsuba (approved):

PR #277, opened 2020-12-14
Add logic for delete contexts by criteria

vkatsuba:
**Extend delete contexts REST API and API for delete contexts contexts by criteria:**   

-  `/api/v1/contexts?version=VERSION&imsi=IMSI&pfcp=PFCP`  

____
**REST API Example:** 

- `/api/v1/contexts?version=v1&imsi=111111111111111&upf-fqdn=prox01.epc.mnc001.mcc001.3gppnetwork.org`  
- `/api/v1/contexts?upf-fqdn=prox01.epc.mnc001.mcc001.3gppnetwork.org`  

____
**API Example:**   
```erlang
ergw_api:delete_contexts(#{
    version => v1,
    'upf-fqdn' => "prox01.epc.mnc001.mcc001.3gppnetwork.org", 
    imsi => <<"111111111111111">>
}).
```
**OR**
```erlang
ergw_api:delete_contexts(#{'upf-fqdn' => "prox01.epc.mnc001.mcc001.3gppnetwork.org"}).
```
____
Query parameters:
| Name   |      Type      |  Description |
|-----------|:---------------:|-----------------:|
| imsi                 | integer | IMSI |
| upf-fqdn          | string | UPF FQDN |
| version            | string | Version |

### DIFF
--- a/priv/static/specs/http_api_spec.yaml
+++ b/priv/static/specs/http_api_spec.yaml
@@ -55,6 +55,25 @@ paths:
   /contexts:
     delete:
       summary: 'delete all contexts'
+      parameters:
+        - name: version
+          in: query
+          required: false
+          type: string
+          description: 'Version'
+          default: 'v1'
+        - name: imsi
+          in: query
+          required: false
+          type: integer
+          description: 'International Mobile Subscriber Identity'
+          default: '111111111111111'
+        - name: upf-fqdn
+          in: query
+          required: false
+          type: string
+          description: 'The UPF FQDN will be used as regex'
+          default: 'prox01.epc.mnc001.mcc001.3gppnetwork.org'
       responses:
         '200':
           description: return count of contexts
--- a/src/ergw_api.erl
+++ b/src/ergw_api.erl
@@ -47,6 +47,10 @@ delete_contexts(all) ->
     lists:foreach(fun(Context) ->
         gtp_context:trigger_delete_context(Context)
     end, contexts(all));
+delete_contexts(#{} = Criteria) ->
+    lists:foreach(fun(Context) ->
+        delete_context_by_criteria(Criteria, gtp_context:info(Context), Context)
+    end, contexts(all));
 delete_contexts(Count) ->
     delete_contexts(contexts(all), Count).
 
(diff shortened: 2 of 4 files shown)


priv/static/specs/http_api_spec.yaml
```diff
@@ -55,6 +55,25 @@ paths:
...
+          type: string
+          description: 'Version'
+          default: 'v1'
+        - name: imsi
+          in: query
+          required: false
+          type: integer
+          description: 'International Mobile Subscriber Identity'
+          default: '111111111111111'
+        - name: pfcp
```
RoadRunnr, line 71: The parameter names makes no sense IMHO. PFCP is a protocol, the entity speaking that protocol is a UPF or in 5G terms "a NF Instance of type `UPF`".

The parameter name should therefor IMHO be `upf-fqdn`.

When a 5G compatible naming is used, then using two arguments, `target-nf-type`, and `target-nf-instance-id` or `target-nf-fqdn` (see TS 29.510, Table 6.2.3.2.3.1-1 for example).

NOTE: In 5G nf-instance-id is a FQDN, but does not need to match the FQDN for instance selection. Therefor nf-fqdn and fn-instance-id could differ.

priv/static/specs/http_api_spec.yaml
```diff
@@ -55,6 +55,25 @@ paths:
...
+          in: query
+          required: false
+          type: string
+          description: 'PFCP'
+          default: 'topon.sx.prox01.epc.mnc001.mcc001.3gppnetwork.org'
```
RoadRunnr, line 76: The example might be misleading. The FQDN represents the name of a specific interface on an UPF instance, the FQDN of the UPF instance itself would be different. In this case the UPF instance name is most likely `prox01.epc.mnc001.mcc001.3gppnetwork.org`

priv/static/specs/http_api_spec.yaml
RoadRunnr: One more thing, the UPF node id is currently missing from the PFCP interaction. This is a oversight that should be fixed soonish (a PR in the UPG project is already pending).

src/http_api_handler.erl
```diff
@@ -95,7 +95,16 @@ handle_request(<<"POST">>, _, json, Req, State) ->
     end;
 
 handle_request(<<"DELETE">>, <<"/api/v1/contexts">>, json, Req, State) ->
-    ok = ergw_api:delete_contexts(all),
+    case cowboy_req:parse_qs(Req) of
+        [] ->
+            ergw_api:delete_contexts(all);
+        Qs ->
+            ergw_api:delete_contexts(#{
+                version    => list_to_atom(binary_to_list(proplists:get_value(<<"version">>, Qs, <<>>))),
```
RoadRunnr, line 103: better use `list_to_existing_atom`

src/ergw_api.erl
```diff
@@ -45,6 +47,11 @@ delete_contexts(all) ->
     lists:foreach(fun(Context) ->
         gtp_context:trigger_delete_context(Context)
     end, contexts(all));
+delete_contexts(#{} = Criteria) ->
+    lists:foreach(fun(Pid) ->
+        {_, State} = sys:get_state(Pid),
```
RoadRunnr, line 52: No,

Don't access process state like state. Either filter through the context_reg or when that doesn't have enough information, use a dedicated API on the context to get that information.

priv/static/specs/http_api_spec.yaml
```diff
@@ -55,6 +55,25 @@ paths:
...
+          in: query
+          required: false
+          type: integer
+          description: 'International Mobile Subscriber Identity'
+          default: '111111111111111'
+        - name: upf-fqdn
+          in: query
+          required: false
+          type: string
+          description: 'UPF FQDN'
```
RoadRunnr, line 75: the implementation actually treats upf-fqdn as regex. Might be worth mentioning here.

RoadRunnr (changes requested):

vkatsuba:
Closed due to lack of activity.

PR #279, opened 2020-12-16
replace context key tupels with records

RoadRunnr:
Instead of having to count fields in tuples, records allow us to
specify matches with the more readable record syntax.
They also give us `tagged keys`, that means we can deduce the type
of the key from the record tag.

This was originally part of the stateless work, but will also be
helpful for the cluster registry.

### DIFF
--- a/test/ggsn_SUITE.erl
+++ b/test/ggsn_SUITE.erl
@@ -973,9 +973,10 @@ path_failure() ->
     [{doc, "Check that Create PDP Context works and "
       "that a path failure (Echo timeout) terminates the session"}].
 path_failure(Config) ->
+    CtxKey = #context_key{socket = 'irx', id = {imsi, ?'IMSI', 5}},
     {GtpC, _, _} = create_pdp_context(Config),
 
-    {_Handler, CtxPid} = gtp_context_reg:lookup({'irx', {imsi, ?'IMSI', 5}}),
+    {_Handler, CtxPid} = gtp_context_reg:lookup(CtxKey),
     #{left_tunnel := #tunnel{socket = CSocket}} = gtp_context:info(CtxPid),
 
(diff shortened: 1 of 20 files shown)


vkatsuba (approved):

test/ggsn_SUITE.erl
```diff
@@ -973,9 +973,10 @@ path_failure() ->
     [{doc, "Check that Create PDP Context works and "
       "that a path failure (Echo timeout) terminates the session"}].
 path_failure(Config) ->
+    CtxKey = #socket_key{name = 'irx', key = {imsi, ?'IMSI', 5}},
```
ebengt, line 976: It might be an idea to define 5 to something more informative.

test/ggsn_SUITE.erl
```diff
@@ -1414,10 +1415,11 @@ delete_pdp_context_requested() ->
     [{doc, "Check GGSN initiated Delete PDP Context"}].
 delete_pdp_context_requested(Config) ->
     Cntl = whereis(gtpc_client_server),
+    CtxKey = #socket_key{name = 'irx', key = {imsi, ?'IMSI', 5}},
 
     {GtpC, _, _} = create_pdp_context(Config),
 
-    {_Handler, Server} = gtp_context_reg:lookup({'irx', {imsi, ?'IMSI', 5}}),
+    {_Handler, Server} = gtp_context_reg:lookup(CtxKey),
```
ebengt, line 1422: This (1418 + 1422) happens often enough to maybe merit its own function?

ebengt (approved):
Looks good

RoadRunnr:
updated, key types are better differentiated now.

vkatsuba (approved):

PR #282, opened 2021-01-07
Fix online charging URR link

fholzhauser:
Linking of online charging URRs in the has the side effect of triggering online charging updates in the offline charging periods instead of considering previously granted volume/time. This PR removes the linking of online URRs as a workaround.

### DIFF
--- a/test/ggsn_SUITE.erl
+++ b/test/ggsn_SUITE.erl
@@ -2017,7 +2017,6 @@ simple_ocs(Config) ->
 	     #measurement_method{volum = 1, durat = 1},
 	 reporting_triggers =>
 	     #reporting_triggers{
-		linked_usage_reporting = 1,
 		time_quota = 1,   time_threshold = 1,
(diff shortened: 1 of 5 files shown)


test/ggsn_SUITE.erl
```diff
@@ -2017,7 +2017,7 @@ simple_ocs(Config) ->
 	     #measurement_method{volum = 1, durat = 1},
 	 reporting_triggers =>
 	     #reporting_triggers{
-		linked_usage_reporting = 1,
+		linked_usage_reporting = 0,
```
vkatsuba, line 2020: By default `linked_usage_reporting` has value `0`, I suppose need remove `linked_usage_reporting` from this place. The same in other places in `*_SUITE.erl`.

fholzhauser: updated, thanks

vkatsuba (approved):

RoadRunnr (approved):
I would still prefer a real fix.

PR #286, opened 2021-01-13
Added PFCP Sx association metric

vkatsuba:
Added Sx association and session counters

### DIFF
--- a/src/ergw_prometheus_collector.erl
+++ b/src/ergw_prometheus_collector.erl
@@ -1,0 +1,80 @@
+%% Copyright 2020, Travelping GmbH <info@travelping.com>
+
+%% This program is free software; you can redistribute it and/or
+%% modify it under the terms of the GNU General Public License
+%% as published by the Free Software Foundation; either version
+%% 2 of the License, or (at your option) any later version.
+
+-module(ergw_prometheus_collector).
+-behaviour(prometheus_collector).
+
+-include("include/ergw.hrl").
+-include_lib("prometheus/include/prometheus.hrl").
+
+-export([
+    deregister_cleanup/1,
+    collect_mf/2,
+    collect_metrics/2,
+    gather/0
+]).
+
+-ignore_xref([gather/0, collect_metrics/2]).
+
+-import(prometheus_model_helpers, [create_mf/5, gauge_metric/2]).
+
+-define(METRIC_NAME_PREFIX, "").
+
+-define(METRICS, [{pfcp_sx_association_total, gauge, "Total number of PFCP Sx association"},
+                  {pfcp_sx_context_total, gauge, "Total number of PFCP Sx node session counters"}]).
+
+%%====================================================================
+%% Collector API
+%%====================================================================
+
+deregister_cleanup(_) ->
+    ok.
+
+collect_mf(_Registry, Callback) ->
+    Stats = gather(),
+    [mf(Callback, Metric, Stats) || Metric <- ?METRICS],
+    ok.
+
+mf(Callback, {Name, Type, Help}, Stats) ->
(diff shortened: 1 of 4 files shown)


src/ergw_sx_node_reg.erl
```diff
@@ -86,12 +86,16 @@ handle_death(_Pid, _Reason, State) ->
 handle_call({up, Key, Caps}, _From, State) ->
     case lookup(Key) of
 	{ok, Pid} ->
+	    ergw_prometheus:pfcp_sx_node_association(inc, Key, up),
+	    ergw_prometheus:pfcp_sx_node_association(dec, Key, down),
```
mgumz, line 90: whats the purpose of this inc-dec dance?

vkatsuba: The main reason of this dancing in calculation of total nodes what was `up` and `down`. If we will do only inc of `up` nodes this will be not correct because the same node can `down` at the moment of receiving the metrics, so we need to do inc of `up` metric if node will `down`. If we will need calculate all nodes what was run and try get only `up` nodes this will be also not correct because some f these nodes can be `down` and correct calculation will be `up + down = total`. But if this is not needed - the logic can be removed. Basically this implementation try separate these two different states.

src/ergw_prometheus_collector.erl
```diff
@@ -0,0 +1,85 @@
...
+gather([], Acc) ->
+    Acc;
+gather([{pfcp_sx_association_total = Key, _, _}|T], Acc) ->
+    gather(T, Acc#{Key => gather_sx_association()});
+gather([{pfcp_sx_context_total = Key, _, _}|T], Acc) ->
+    gather(T, Acc#{Key => gather_sx_context()}).
+
+gather_sx_association() ->
+    #{
+        [{state, up}] => maps:size(ergw_sx_node_reg:available())
```
RoadRunnr, line 74: I would prefer if you do the maps:size internally in ergw_sx_node_reg. Add a `count_available` (or similar named method to retrieve the value.

src/ergw_prometheus_collector.erl
```diff
@@ -0,0 +1,85 @@
...
+
+gather_sx_context() ->
+    maps:fold(fun(Name, {Pid, _}, Stats) ->
+        case process_info(Pid, monitored_by) of
+            {monitored_by, Monitors} ->
+                Stats#{[{name, Name}] => length(Monitors)};
+            _ ->
+                Stats
+        end
+    end, #{}, ergw_sx_node_reg:available()).
```
RoadRunnr, line 85: At least the process_info / monitored magic should be done in ergw_sx_node_reg. Exposing this implementaion detail outside of the ergw_sx_node_reg is IMHO not sensible. Iterating over the list of nodes can remain here.

test/ggsn_SUITE.erl
```diff
@@ -2676,6 +2677,15 @@ up_inactivity_timer(Config) ->
...
     ok.
 
+%%--------------------------------------------------------------------
+pfcp_sx_association_metric() ->
+    [{doc, "Test expiry of PFCP Sx association and session counters"}].
+pfcp_sx_association_metric(_Config) ->
+    ?equal(ok, prometheus_registry:register_collector(ergw, ergw_prometheus_collector)),
+    Metrics = prometheus_text_format:format(ergw),
+    ?equal(true, is_match(re:run(Metrics, "pfcp_sx_association_total"))),
+    ?equal(true, is_match(re:run(Metrics, "pfcp_sx_context_total"))).
```
RoadRunnr, line 2687: is `?match` instead of `?equal` and drop the `is_match` helper.

RoadRunnr (approved):

PR #288, opened 2021-01-19
add GH CI workflow

RoadRunnr:


fholzhauser (approved):

vkatsuba (commented):
Looks like if Travis will removed we also don't need `test/env.sh`, `rebar.config.script`.

RoadRunnr:
> Looks like if Travis will removed we also don't need `test/env.sh`, `rebar.config.script`.

Good points:
* the SLOW_TEST setting from the test/env.sh should be integrated into the GH action
* the coveralls_job_id from the rebar.config.script needs to be changed to use a GH id.

vkatsuba:
> Good points:
>     * the SLOW_TEST setting from the test/env.sh should be integrated into the GH action
>     * the coveralls_job_id from the rebar.config.script needs to be changed to use a GH id.
Done.

vkatsuba:
> Good points: 
>     * the SLOW_TEST setting from the test/env.sh should be integrated into the GH action
>     * the coveralls_job_id from the rebar.config.script needs to be changed to use a GH id.

Done.

.github/workflows/main.yml
```diff
@@ -58,3 +65,33 @@ jobs:
...
+        fields: repo, message, ref, commit, author, action, eventName, workflow, job, took
+        status: ${{ needs.test.result }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
+        MATRIX_CONTEXT: ${{ toJson(matrix) }}
+
+  finish:
+    needs: test
+    runs-on: ubuntu-20.04
```
RoadRunnr, line 89: finish is not called when on the test fails, there should be a `if: always()` here.

vkatsuba: Yes, but I was think that the testes must pass before call finish. Can you confirm that this behavior(`if: always()`) is also expected for other repo?

RoadRunnr: yes, I think so. Otherwise we are not going to get the coverage for failed tests.

vkatsuba: Thanks. Done.

RoadRunnr:
looks ok. However, the README.md has this:

> Docker images are build by travis-ci and pushed to hub.docker.com, and by gitlab.com and pushed to quay.io.

That should be adjusted as well. Without travis, images are no longer pushed to docker hub.

vkatsuba:
> That should be adjusted as well. Without travis, images are no longer pushed to docker hub.

Yep, I want to play around with [Publishing Docker images](https://docs.github.com/en/actions/guides/publishing-docker-images), before merge

.github/workflows/main.yml
```diff
@@ -0,0 +1,105 @@
...
+           ip addr add fd96:dcd2:efdb:41c3::20/64 dev lo
+           ip addr add fd96:dcd2:efdb:41c3::30/64 dev lo
+           ip addr add fd96:dcd2:efdb:41c3::40/64 dev lo
+           ip addr add fd96:dcd2:efdb:41c3::50/64 dev lo
+    - name: Run tests
+      env:
+        CI_DNS_SERVER: 172.18.0.1
+        CI_RUN_SLOW_TESTS: true
+      run: |
+           docker run -d --rm \
```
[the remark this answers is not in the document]
vkatsuba: Discussed.

PR #297, opened 2021-01-25
Fix Swagger GET metrics

vkatsuba:
Fixed few issues:
- The `basePath` variable was set base path(`/api/v1`) for each REST API, while the other REST API can use path without `/api/v1`
- The default accept type `application/json` is not suit for REST API `/metrics`

### DIFF
--- a/priv/static/specs/http_api_spec.yaml
+++ b/priv/static/specs/http_api_spec.yaml
@@ -6,9 +6,9 @@ info:
   license:
     name: "GPL 2.0"
     url: "https://www.gnu.org/licenses/#GPL"
-basePath: /api/v1
+basePath: /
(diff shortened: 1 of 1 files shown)


RoadRunnr:
/metrics might not really belong into swagger / API definition. Is not a RPC API, but rather a special function that just happen to be on the same server.

Removing /metric from the API specification might be the better approach.

@fholzhauser @mgumz opinions?

mgumz:
@RoadRunnr /metrics: yes, should be outside of an api.

vkatsuba:
Closed PR and opened https://github.com/travelping/ergw/pull/313 for `/metrics` from Swagger definitions.

PR #304, opened 2021-01-27
multiarch build support on Travis and GitLab-CI

RoadRunnr:


.gitlab-ci.yml
```diff
@@ -61,46 +37,53 @@ build:otp-23.1:
     - ip addr add fd96:dcd2:efdb:41c3::30/64 dev lo
     - ip addr add fd96:dcd2:efdb:41c3::40/64 dev lo
     - ip addr add fd96:dcd2:efdb:41c3::50/64 dev lo
-    - ./rebar3 do xref
-    - ./rebar3 do ct
+    - rebar3 xref
```
vkatsuba, line 40: From where the `rebar3` will install by default in CI? Will this version match the version of the `rebar3` that can be get in `https://s3.amazonaws.com/rebar3/rebar`?

vkatsuba (approved):

.gitlab-ci.yml
RoadRunnr: The Erlang images do contain rebar. In case you want to chech, here is one the Dockerfiles use to build those images: https://github.com/erlang/docker-erlang-otp/blob/master/23/alpine/Dockerfile

fholzhauser (approved):

PR #306, opened 2021-01-28
Implement configurable GTP RTT metrics interval

vkatsuba:


fholzhauser (approved):

src/ergw_app.erl
```diff
@@ -22,10 +22,10 @@
 start(_StartType, _StartArgs) ->
     do([error_m ||
 	   gtp_config:init(),
-	   ergw_prometheus:declare(),
 	   ensure_jobs_queues(),
 	   Pid <- ergw_sup:start_link(),
 	   ergw_config:load_config(setup:get_all_env(ergw)),
+	   ergw_prometheus:declare(),
```
RoadRunnr, line 28: there is a small chance that moving it here is not safe. Needs further checks.

vkatsuba: what if we move `prometheus_histogram:declare([{name, gtp_path_rtt_milliseconds ...` into other exported function and will call it after option validation?

RoadRunnr: problem is that ergw_config:load_config starts the sockets. The moment the socket is up, it could try to change a metric that might not declared at that point.

vkatsuba: this changes will reverted after merge https://github.com/travelping/ergw/pull/310

vkatsuba: Done. Thanks.

RoadRunnr (approved):

PR #315, opened 2021-02-02
Implement configurable GTP path ICMP error behaviour

vkatsuba:


fholzhauser (approved):

src/gtp_path.erl
```diff
@@ -318,7 +325,12 @@ handle_event(cast,{handle_response, echo_request, _, _Msg}, State, Data) ->
     path_restart(undefined, peer_state(down, State), Data, []);
 
 handle_event(cast, down, State, Data) ->
-    path_restart(undefined, peer_state(down, State), Data, []);
+    case Data of
+        #{icmp_error_handling := ignore} ->
+            keep_state_and_data;
+        _ ->
+            path_restart(undefined, peer_state(down, State), Data, [])
+    end;
```
RoadRunnr, line 333: this is wrong place to check for the flag. I needs to go close to the ICMP handling.

RoadRunnr (changes requested):

src/gtp_path.erl
RoadRunnr: just rechecked and gtp_path:down/2 is only used from the ICMP logic. Still, unconditionally use a test for ICMP on an interface that has a very generic name is not good.

So, I would suggest to add a `Source` field to the API, forward that value in cast, populate that with the value `icmp` in the socket module and combine the icmp_error check with a check for the icmp value.

RoadRunnr: or alternatively, rename `down` to something else, e.g. `icmp_error`

vkatsuba: Please, re review.

src/ergw_gtp_c_socket.erl
```diff
@@ -143,6 +144,8 @@ init(#{name := Name, ip := IP, burst_size := BurstSize,
        send_port := SendPort} = SocketOpts) ->
     process_flag(trap_exit, true),
 
+    {ok, #{icmp_error_handling := Act}} = application:get_env(ergw, path_management),
```
RoadRunnr, line 147: My previous comments might have been a bit confusing, they were meant to give alternatives not that all of them have to used at once.

So, the icmp_error renaming is fine, but:
a) either continue checking the  icmp_error_handling in the socket logic, if you choose that, then the config options has to go into the socket options
b) check the option only in the gtp_path logic, then the option can in stay path_management, no check need to be performed in the socket logic.

I would prefer variant b).

vkatsuba: Thanks for clarification, done.

RoadRunnr (approved):
LGTM

PR #323, opened 2021-02-18
Return SGW TEID in GTP session/context creation response header

vkatsuba:


test/ergw_ggsn_test_lib.erl
```diff
@@ -431,14 +431,17 @@ validate_seq_no(#gtp{seq_no = SeqNo}, #gtpc{seq_no = GtpCSeqNo}) ->
 validate_seq_no(#gtp{seq_no = SeqNo}, ExpectedSeqNo) ->
     ?equal(ExpectedSeqNo, SeqNo).
 
+validate_teid(#gtp{tei = 0} , #gtpc{}) ->
+    ok;
```
RoadRunnr, line 435: don't do that. Testing for zero TEID should remain explicit in the places where such a TEID is indeed expected.

test/ergw_ggsn_test_lib.erl
```diff
@@ -431,14 +431,17 @@ validate_seq_no(#gtp{seq_no = SeqNo}, #gtpc{seq_no = GtpCSeqNo}) ->
...
 validate_teid(#gtp{tei = TEID}, #gtpc{local_control_tei = GtpCTEID}) ->
     ?equal(GtpCTEID, TEID);
 validate_teid(#gtp{tei = TEID}, ExpectedTEID) ->
     ?equal(ExpectedTEID, TEID).
 
+-define(rec_info(T,R),lists:zip(record_info(fields,T),tl(tuple_to_list(R)))).
```
RoadRunnr, line 441: This is unused.

src/gtp_context.erl
```diff
@@ -745,9 +745,9 @@ get_handler_if(Socket, #gtp{version = v2} = Msg) ->
     gtp_v2_c:get_handler(Socket, Msg).
 
 find_sender_teid(#gtp{version = v1} = Msg) ->
-    gtp_v1_c:find_sender_teid(Msg);
+    gtp_v1_c:find_sender_teid(gtp_packet:decode_ies(Msg));
```
RoadRunnr, line 748: This will lead to decoding the IEs twice. The IE decoding is delayed by intention. It is expensive and should not happen in the socket process.
Most errors that require a TEID in the response will happen in places where the IEs are already decoded. Change those places to return the sender TEID in their error response.

RoadRunnr (changes requested):

src/ergw_gtp_c_socket.erl
```diff
@@ -453,7 +453,7 @@ handle_err_input(Socket, State) ->
     State.
 
 handle_message(ArrivalTS, Src, IP, Port, Data, #state{gtp_socket = Socket} = State0) ->
-    try gtp_packet:decode(Data, #{ies => binary}) of
+    try gtp_packet:decode(Data, #{ies => map}) of
```
RoadRunnr, line 456: decoding to a map is too soon. This should only happen in the handler process, not in the socket process

RoadRunnr (changes requested):

RoadRunnr:
This change tries to do too much.

Returning a response with TEID = 0 when the cause is `context_not_found` is perfectly legal and explicitly documented for GTPv2.
All the other error paths are mostly correct with setting the TEID in error responses.

The only problem is in the place that is mentioned in the actual bug #324 , that is when accept_new == false.

I'll open a more limited PR soonish.

PR #348, opened 2021-03-23
Change default value of 'idle_timeout'

vkatsuba:
Start use `infinity` value in `idle_timeout`
Motivation:
- by default - this functionality is not needed
- in specific scenarios we should set the value to 12h


RoadRunnr (changes requested):
Can you explain where that comes from?
Idle timeout was explicitly added to remove paths that are no longer in use, why has that changed now?

Issue #346, opened 2021-03-19
wrong PFCP control flow during HSS Initiated Subscribed QoS Modification


RoadRunnr:
HSS Initiated Subscribed QoS Modification is defined in TS 23.401 clause 5.4.2.2 and updated for CUPS in TS 23.214 clause 6.3.1.7.

The resulting control flow should invoke a PFCP session modification only after the Update Bearer Response (step in 23.401) has been received.

We currently trigger a PFCP usage query on receiving the modification command and trigger another PFCP usage query when getting the Update Bearer Response. Since QoS is currently not supported, updating the bearers in the UPF is not needed, but the first usage query is not needed at all and the second query is only needed when the QoS settings have indeed been changed.

A race between the session termination when the update bearer procedure returns a failure and the first usage query causes random test case failures in pgw_suite.modify_bearer_command_congestion.

Failure is update bearer requests are rare, so this bug has only medium priority.

PR #354, opened 2021-03-30
refactor application split and with it also config handling

RoadRunnr:
# STATUS

Passes most tests suites. TODO are in state that the work can distributed across several developers (set TODOs)

# CHANGES

- split ergw into core, cluster and top level application functions
- core is only configured through API and is intended be used as library
- top level is responsible for read config and applying it to cluster, core and aaa
- non core API handlers (mostly HTTP) are moved to top level. This might be revised and the handler might move to extra apps later
- repository layout is currently using a Erlang multi application approach, this might be revised later

# TODOs

- [ ] rework all documentation pieces
- [ ] review changes for consistency
- [x] check and if needed fix the pgw_dist_SUITE
- [ ] add the missing aaa key to the JSON schema for the config
- [ ] enhance the JSON schema (e.g. add 'required' for mandatory items, add additional range restrictions to integers, ...)
- [x] (optional) write a config converter


RoadRunnr:
test cases are failing in pgw_SUITEs `modify_bearer_command_congestion` case. That problem is already captured in #346 and unrelated to this PRs.

vkatsuba (approved):

PR #355, opened 2021-04-07
DNS: add negative caching for empty responses

RoadRunnr:
this is stacked on top of #354


vkatsuba (commented):
Do we need also cherry pick `DNS: add negative caching for empty responses` to some new branch which based on `2.8.9` release?

RoadRunnr:
> Do we need also cherry pick `DNS: add negative caching for empty responses` to some new branch which based on `2.8.9` release?

I don't think that makes sense, let 2.8.x use the old cache and use the new one for master

vkatsuba (approved):

PR #361, opened 2021-04-13
add experimental location dependent calculation to Tariff-Time-Change

RoadRunnr:
This feature should be treated as experimental even when merged.

The preconfigure Tariff-Time in charging profile now has a location
attribute. That location attribute is use to interpret the time a
local time and translate it to UTC for use as a Monitoring Time
in PFCP.

As location name all names from the olsen timezone database are
valid. To specify a UTC time use "Etc/UTC" as location.

This requires also a change to "Tariff-Time" declaration in the config. In the JSON config, it now needs to be something like:

```
"Tariff-Time" : {
    "Local-Tariff-Time": "00:00",
    "Location": "Europe/Berlin"
}
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
            apk --no-cache upgrade
            apk --no-cache add docker gcc git libc-dev libc-utils libgcc linux-headers make bash \
                               musl-dev musl-utils ncurses-dev pcre2 pkgconf scanelf wget zlib \
-                              yaml-dev
+                              yaml-dev tzdata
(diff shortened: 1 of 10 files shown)


ebengt (commented):
Previously the problem with changing tariff-time-change was that the entity producing the report at the configured time (u-node) would have the old value for existing sessions.
(Or I misunderstood the problem)
Does this pull request handle that problem?

vkatsuba (approved):

hudymi:
@vkatsuba which release contains this feature?

vkatsuba:
@hudymi `3.1.0-rc3 `

PR #363, opened 2021-04-14
move config parts into the global state (raft) handler

RoadRunnr:


apps/ergw_core/src/ergw_core_config.erl
```diff
@@ -36,7 +36,21 @@ to_map(L) when is_list(L) ->
...
+	true  -> get_global_key(Query, Default);
+	false -> get_local_key(Query, Default)
+    end.
+
+get_global_key(Query, Default) ->
+    case ergw_global:find(Query) of
+	{ok, Value} ->
+	    {ok, Value};
+	_ ->
+	    {ok, Default}
```
vkatsuba, line 50: This function will return value in any case, maybe make sense return only value instead of tuple, eg like in [get_env/3](https://erlang.org/doc/apps/kernel/application.html#get_env-3)?

apps/ergw_core/src/ergw_core.erl
```diff
@@ -77,16 +77,16 @@ is_running() ->
 
 %% get global PLMN Id (aka MCC/MNC)
 get_plmn_id() ->
-    [{config, plmn_id, #{mcc := MCC, mnc := MNC}}] = ets:lookup(?SERVER, plmn_id),
+    {ok, #{mcc := MCC, mnc := MNC}} = ergw_core_config:get([plmn_id], undefined),
```
vkatsuba, line 80: Looks like will be problem if will return `{ok, undefined}` default value

apps/ergw_core/src/ergw_core_config.erl
vkatsuba, line 50: It would be informative if in the log will provided information that was provided the default value was taken.

apps/ergw_core/src/ergw_core_config.erl
```diff
@@ -56,7 +70,26 @@ get(_, _, _) ->
     undefined.
 
 put(Key, Val) ->
-    ok = application:set_env(ergw_core, Key, Val).
+    case is_global_key(Key) of
+	true ->
+	    {ok, _} = ergw_global:put(Key, Val),
```
vkatsuba, line 75: The `ergw_global:put/2` can return also `timeout` - maybe make sense handle it?

apps/ergw_core/src/ergw_core.erl
RoadRunnr: yes, that is intended. If this function is called before ergw is properly configure, then it should crash.

apps/ergw_core/src/ergw_core_config.erl
RoadRunnr: why? there could be thousands of those case, that would only spam the log.

RoadRunnr: the {ok, _} return is preparation of also returning error later, e.g. it might be better to return an error when the raft handler fails.

apps/ergw_core/src/ergw_core_config.erl
RoadRunnr: in that case the system should fail anyway. Remember, in Erlang you code the ok case and let it crash on errors.

vkatsuba (approved):

PR #366, opened 2021-04-21
match request content when checking for resents

RoadRunnr:
3GPP specification are silent on how to check duplicate requests.
The only statement is that resent requests are identical to the
initial requests.

For GTPv1 the sequence number range is only 16 bit and will roll over
within the retransmistion window for rates of about 1000 req/s per peer.

Artificial traffic generators and naive GTP probes might produce
sequence number collisions even sooner.

To avoid those collision, make a resonable safe and fast hash of the
payload part of key to detect retransmissions.


apps/ergw_core/src/ergw_gtp_c_socket.erl
```diff
@@ -174,8 +174,8 @@ init({Name, #{ip := IP, vrf := VRF,
 	       v1_seq_no = 0,
 	       v2_seq_no = 0,
 	       pending = gb_trees:empty(),
-	       requests = ergw_cache:new(?T3 * 4, requests),
-	       responses = ergw_cache:new(?CACHE_TIMEOUT, responses),
+	       requests = ergw_cache:new(100, requests),
```
vkatsuba, line 177: Do we need also set `100` into [ergw_sx_socket.erl#L168](https://github.com/travelping/ergw/blob/master/apps/ergw_core/src/ergw_sx_socket.erl#L168)?

apps/ergw_core/src/ergw_gtp_c_socket.erl
```diff
@@ -174,8 +174,8 @@ init({Name, #{ip := IP, vrf := VRF,
 	       v1_seq_no = 0,
 	       v2_seq_no = 0,
 	       pending = gb_trees:empty(),
-	       requests = ergw_cache:new(?T3 * 4, requests),
-	       responses = ergw_cache:new(?CACHE_TIMEOUT, responses),
```
vkatsuba, line 178: Looks like the `CACHE_TIMEOUT` macros is not used anymore in current module. Please delete this non use macros.

vkatsuba (approved):

PR #367, opened 2021-04-22
Remove GTP Path RTT milliseconds metric for release 2.8.11

vkatsuba:


NEWS.md
```diff
@@ -1,5 +1,11 @@
 erGW - 3GPP GGSN and PDN-GW in Erlang
 =====================================
+
+Version 2.8.10 - 22 April 2021
+-------------------------------
+**Improvements** :bulb:
```
fholzhauser, line 6: Could you please change this from Improvement to Hotfix ? This change is not really improving functionality, but a short term fix for a problem with scraping when there are many GTP peers and metrics, with this RTT metric as the biggest contributor, become too big to export for prometheus.erl.

vkatsuba:
Closed due to https://github.com/travelping/ergw/pull/367

PR #368, opened 2021-04-22
Remove GTP Path RTT milliseconds metric

vkatsuba:


fholzhauser:
This change is not needed on master, we'll come up with a different solution on long term. I'm closing this PR.

PR #371, opened 2021-04-26
Start use fork of prometheus

vkatsuba:
Started using fork of `prometheus` with speedup fix for `master`


RoadRunnr (changes requested):
upstream has accepted all our patches, but no new release yet.
Using my fork is therefore not longer required.

vkatsuba:
Closed. Will be opened new PR after `prometheus` release.

PR #377, opened 2021-05-04
Change 'normal' termination reason

vkatsuba:
Changed 'normal' termination reason to expected 'cp_inactivity_timeout'

### DIFF
--- a/apps/ergw_core/src/gtp_context.erl
+++ b/apps/ergw_core/src/gtp_context.erl
@@ -612,7 +612,7 @@ handle_event(info, {'DOWN', _MonitorRef, Type, Pid, _Info}, State,
     {next_state, shutdown, Data};
 
 handle_event({timeout, context_idle}, stop_session, State, Data) ->
-    delete_context(undefined, normal, State, Data);
+    delete_context(undefined, cp_inactivity_timeout, State, Data);
 
 handle_event(Type, Content, State, #{interface := Interface} = Data) ->
     ?LOG(debug, "~w: handle_event: (~p, ~p, ~p)",


ebengt:
The failed Gitlab in CI could be what I experienced last week. Random failure with
Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?
Solved by adding version to docker image. Example:
    image: docker:19.03.8
    services:
        - docker:19.03.8-dind

apps/ergw_core/src/gtp_context.erl
```diff
@@ -612,7 +612,7 @@ handle_event(info, {'DOWN', _MonitorRef, Type, Pid, _Info}, State,
     {next_state, shutdown, Data};
 
 handle_event({timeout, context_idle}, stop_session, State, Data) ->
-    delete_context(undefined, normal, State, Data);
+    delete_context(undefined, cp_inactivity_timeout, State, Data);
```
ebengt, line 615: Who wants/expects
cp_inactivity_timeout
It is not part of ergw.

ebengt (approved):

vkatsuba:
> The failed Gitlab in CI could be what I experienced last week. Random failure with
> 
> Cannot connect to the Docker daemon at tcp://localhost:2375. Is the docker daemon running?
> 
> Solved by adding version to docker image. Example:
> 
>     image: docker:19.03.8
> 
>     services:
> 
>         - docker:19.03.8-dind
> 
> 

Thanks but as I know that the GitLab CI should be migrated to GitHub actions soon, but maybe this fix also will be need in GH actions.

apps/ergw_core/src/gtp_context.erl
vkatsuba: You are right it is not a part of erGW - this is a part of erGW-AAA and will used for termination cause mapping.

PR #378, opened 2021-05-07
Start use quay.io/travelping/alpine-erlang:23.3.3

vkatsuba:


ebengt (commented):
Which erlang:23.3.3 ? Is it Travelping version with old SCTP libraries?

docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # -- build-environment --
 # see https://docs.docker.com/engine/userguide/eng-image/multistage-build/
 
-FROM erlang:23.2.7.0-alpine AS build-env
+FROM erlang:23.3.3-alpine AS build-env
```
RoadRunnr, line 4:
```suggestion
FROM quay.io/travelping/alpine-erlang:23.3.3 AS build-env
```

RoadRunnr (changes requested):

vkatsuba:
> Which erlang:23.3.3 ? Is it Travelping version with old SCTP libraries?

Updated.

PR #381, opened 2021-05-11
Migrate publish to 'quay.io' from GL to GH

vkatsuba:


RoadRunnr (commented):
Why does this have to be so different from the docker.yaml. Sure, using the RedHat toolchain is new and existing, but is there a real benefit that justifies to have two different ways of building images?

RoadRunnr (changes requested):
aside from other comment, builds that are no triggered by tags or releases should have a expiry on quay. There is no point on keeping builds for every commit indefinitely.

vkatsuba:
> Why does this have to be so different from the docker.yaml. Sure, using the RedHat toolchain is new and exiting, but is there a real benefit that justifies to have two different ways of building images?
> 
> Also, the gitlab action and the docker action also build for arm64. We should consider if we want to keep that or not.

I was try use `docker.yaml` as template but I was get errors like `buildx call failed with: error: failed to solve: rpc error: code = Unknown desc = server message: insufficient_scope: authorization failed` [CI Result](https://github.com/travelping/ergw/runs/2558418850?check_suite_focus=true). I was also see that the `docker_meta` also not totally suit for publish to `quay.io`.After some research was found this approach and used.

vkatsuba:
> aside from other comment, builds that are no triggered by tags or releases should have a expiry on quay. There is no point on keeping builds for every commit indefinitely.

Yes, this point I want to discuss better if you not mind, could you please provide full list of rules for build images(eg, expire/naming/tagging/rewriting images). Eg, do we need have `latest:branch_name` for latest build etc.?

vkatsuba:
Closed due to alternative version of #383.

PR #386, opened 2021-05-25
Start use quay.io/travelping/alpine-erlang:23.3 for 3.X

vkatsuba:


RoadRunnr (changes requested):
libgcc and libstdc++ are not need in final image for OTP-23.3

vkatsuba:
> libgcc and libstdc++ are not need in final image for OTP-23.3

Done, thanks.

PR #394, opened 2021-05-31
lowercase APN before further processing it

RoadRunnr:
The 3GPP specification do not spell it out but seem to assume that
APN names are used in all places as case insensitive.
Lowercase all APN names from GTP before using it further.

### DIFF
--- a/apps/ergw_core/src/ergw_node_selection.erl
+++ b/apps/ergw_core/src/ergw_node_selection.erl
@@ -87,6 +87,9 @@ candidates(Name, Services, NodeSelection) ->
 	    L
     end.
 
+lowercase_apn([H|_] = APN) when is_binary(H) ->
+    [string:lowercase(X) || X <- APN].
+
 expand_apn_plmn(IMSI) when is_binary(IMSI) ->
     {MCC, MNC, _} = itu_e212:split_imsi(IMSI),
     {MCC, MNC};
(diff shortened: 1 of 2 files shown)


apps/ergw_core/src/ergw_node_selection.erl
```diff
@@ -87,6 +87,9 @@ candidates(Name, Services, NodeSelection) ->
 	    L
     end.
 
+lowercase_apn([H|_] = APN) when is_binary(H) ->
+    [string:lowercase(X) || X <- APN].
```
vkatsuba, line 91: Maybe make sense try use `cowboy_bstr:to_lower/1`

fholzhauser (approved):

apps/ergw_core/src/ergw_node_selection.erl
fholzhauser: probably not the best spot to introduce a dependency on cowboy

RoadRunnr: No,
a) cowboy is not dependency of ergw_core,
b) what does `cowboy_bstr:to_lower/1` better than `string:lowercase/1`

vkatsuba:
```erlang
string:lowercase(<<"привет">>).
<<"?@825b">>
```
bad converting if not utf8.

RoadRunnr: those characters are not permitted in APN names anyway.

yes, we could validate the input a bit further, but failing on this will only make the handler for the current request fail. Since the input is invalid, that is ok.

RoadRunnr: if you truly insisted on using that string as APN, it would need to be encoded as `xn--b1agh1afp` to be valid. UI might then choose to actually display that as unicode string.

vkatsuba: I see, thanks.

PR #404, opened 2021-06-03
add PSDBU part or the EPFAR feature

RoadRunnr:
Support `PFCP Session Deleted By the UP function` procedure that is part
of the EPFAR (Enhanced PFCP Association Release) feature.
The actually release procedure is not supported, but the "Deleted By the UPF"
can already be used to signal session deletion in the UPF

### DIFF
--- a/apps/ergw_core/src/gtp_context.erl
+++ b/apps/ergw_core/src/gtp_context.erl
@@ -387,6 +387,23 @@ handle_event({call, From},
 	    {next_state, shutdown, Data, [{reply, From, {ok, PCtx}}]}
     end;
 
+%% PFCP Session Deleted By the UP function
+handle_event({call, From},
+	     {sx, #pfcp{type = session_report_request,
+			ie = #{pfcpsrreq_flags := #pfcpsrreq_flags{psdbu = 1},
+			       report_type := ReportType} = IEs}},
+	     _State, #{pfcp := PCtx} = Data0) ->
+    TermCause =
+	case ReportType of
+	    #report_type{upir = 1} ->
+		up_inactivity_timeout;
+	    _ ->
+		deleted_by_upf
+	end,
+    UsageReport = maps:get(usage_report_srr, IEs, undefined),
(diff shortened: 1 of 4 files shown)


apps/ergw_core/src/gtp_context.erl
```diff
@@ -387,6 +387,23 @@ handle_event({call, From},
...
+	     {sx, #pfcp{type = session_report_request,
+			ie = #{pfcpsrreq_flags := #pfcpsrreq_flags{psdbu = 1},
+			       report_type := ReportType} = IEs}},
+	     _State, #{pfcp := PCtx} = Data0) ->
+    TermCause =
+	case ReportType of
+	    #report_type{upir = 1} ->
+		up_inactivity_timeout;
+	    _ ->
+		deleted_by_upf
```
vkatsuba, line 401: In `erGW-AAA` we don't handle/mapping `deleted_by_upf` termination cause. Do we need also change `erGW-AAA`? If yes - what number of `deleted_by_upf` should be mapped?

RoadRunnr: 1. ergw_aaa should use a default mapping for unknown (new) causes. That it doesn't do that is IMHO a bug and should be fixed
2. [NAS Error](https://datatracker.ietf.org/doc/html/rfc2866#section-5.10) would IMHO be a reasonable code

vkatsuba: What number should be for DIAMETER https://tools.ietf.org/html/rfc3588#section-8.15?

RoadRunnr: LetMeGoogleThatForYou....

The one assigned by [IANA](https://www.iana.org/assignments/aaa-parameters/aaa-parameters.xhtml#aaa-parameters-16), of course.
The values can also be found in the dictionary: https://github.com/travelping/ergw_aaa/blob/master/dia/diameter_rfc4005_nasreq.dia#L718-L741

vkatsuba: :+1:

fholzhauser (approved):

vkatsuba (approved):

PR #411, opened 2021-06-30
Add ergw_sbi_client

vkatsuba:
* Create new application ergw_sbi_client
* Include ergw_sbi_client into ergw
* Update ergw for validation of ergw_sbi_client
* Include ergw_sbi_client into ergw_core


mgumz:
@vkatsuba what parts of the functionality is missing?

vkatsuba:
> @vkatsuba what parts of the functionality is missing?

Missed the part for generate JSON body request for send it by API. It is still not clear how to correctly form the JSON, from what data, and how we can use response inside of `erGW`. But so far as I know, the @RoadRunnr was start implement something for it or in separate PR or so. So, I suppose he have more info about this, cuz I was know about it only 5 min ago.

RoadRunnr:
limitations, restrictions, requirements and deviations from the API specification:
* only the fqdn return type of the API is supported
* UPF nodes that can be selected need to be configured in the upf_nodes / entries section with a "connect":true setting
* UPGs need to run with the upf specification release 16 setting
* FQDNs in the upf_nodes / entries section and the API return have to match verbatim, no expansions will be performed on them all location information (CGI, SAI, LAI, ...) are optional (even when then API specification has some of them as mandatory), this might no change for the final version, as we do not receive those values in some settings
* QoS-Information is a structured type instead of a string (we might want to keep this for the final version and change the API specification accordingly)

fholzhauser (approved):

PR #412, opened 2021-07-05
enhance VRF and node selection

RoadRunnr:
* rewrite UPF selection and reselection
* enhance VRF selecting in TDF mode to consider NAT pools


RoadRunnr:
there is one TBD left:
> if the old node is also eligible for the new values then we should keep that node.

Does anyone feel that this is important enough to delay merging this PR?

fholzhauser (approved):

vkatsuba (approved):

PR #421, opened 2021-07-30
Fix GTP path echo timer reset

fholzhauser:
This PR would fix the issue that short sessions on a right side path (e.g. with GSN path change on create session response) resets the scheduled echo and can cause lack of path management in case only such sessions are done on a path.

- the gtp path sends an echo immediately when the path starts, speeding up the detection of a broken destination
- the echoes are scheduled relative to the last sent echo which prevents that path management stops due to rescheduling the echos just with the interval at each change

### DIFF
--- a/src/gtp_path.erl
+++ b/src/gtp_path.erl
@@ -41,10 +41,13 @@
 	       contexts :: non_neg_integer()
 	      }).
 
-%% echo_timer is the status of the echo send to the remote peer
+%% echo_is the status of the echo send to the remote peer
+%% - {stopped, LastEchoTs} : no echo idle, with last echo request sent
+%% - {idle, LastEchoTs} : next echo idle with the timestamp of the last echo request sent
+%% - {reference, LastEchoTs} : echo waiting for response with the timestamp of the echo request sent
 -record(state, {peer       :: #peer{},                     %% State of remote peer
 		recovery   :: 'undefined' | non_neg_integer(),
-		echo       :: 'stopped' | 'echo_to_send' | 'awaiting_response'}).
+		echo       :: {'stopped', non_neg_integer()} | {'idle', non_neg_integer()} | {reference(), non_neg_integer()}}). 
 
 %%%===================================================================
 %%% API
@@ -208,8 +211,8 @@ init([#socket{name = SocketName} = Socket, Version, RemoteIP, Args]) ->
     RegKey = {SocketName, Version, RemoteIP},
     gtp_path_reg:register(RegKey, up),
 
-    State = #state{peer = #peer{state = up, contexts = 0},
-		   echo = stopped},
+    State0 = #state{peer = #peer{state = up, contexts = 0},
+		   echo = {stopped, 0}},
 
     Data0 = maps:with([t3, n3, echo,
 		       idle_timeout, idle_echo,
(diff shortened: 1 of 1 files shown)


src/gtp_path.erl
```diff
@@ -563,7 +576,7 @@ send_echo_request(State, #{socket := Socket, handler := Handler, ip := DstIP,
     Ref = erlang:make_ref(),
     CbInfo = {?MODULE, handle_response, [self(), echo_request, Ref]},
     ergw_gtp_c_socket:send_request(Socket, any, DstIP, ?GTP1c_PORT, T3, N3, Msg, CbInfo),
-    State#state{echo = Ref}.
+    State#state{echo = {Ref, os:system_time(millisecond)}}.
```
RoadRunnr, line 579: this should use `erlang:monotic_time`

src/gtp_path.erl
```diff
@@ -403,17 +406,27 @@ enter_state_timeout_action(down, #{down_timeout := Timeout}) when is_integer(Tim
...
-enter_state_echo_action(_, _) ->
+    {{timeout, echo}, next_echo_schedule(EchoInterval, LastEchoTs), start_echo};
+enter_state_echo_action(down, {_, LastEchoTs}, #{down_echo := EchoInterval})
+  when is_integer(EchoInterval) ->
+    {{timeout, echo}, next_echo_schedule(EchoInterval, LastEchoTs), start_echo};
+enter_state_echo_action(_, _, _) ->
     {{timeout, stop_echo}, 0, stop_echo}.
 
+next_echo_schedule(EchoInterval, LastEchoTs) ->
+    TS = os:system_time(millisecond),
```
RoadRunnr, line 422: this should use `erlang:monotic_time`

RoadRunnr (changes requested):
I'm not sure that this change goes into the right direction.
The intention of the Echo logic was always to restart Echo interval when a state change occurs. The reasoning is that a change of state somehow indicates a activity on the peer and that that defines the last know state of the peer. The next Echo will then validate that state (or change it).

The real problem that this PR tries to fix is that pending requests and context will impact the state of the peer. And that is IMHO the real problem, activity triggered by us should never change the peers state, only responses and timeouts from the peer should impact the peer state.

src/gtp_path.erl
fholzhauser: yes, thanks for spotting it

src/gtp_path.erl
fholzhauser: indeed

fholzhauser:
> I'm not sure that this change goes into the right direction.
> The intention of the Echo logic was always to restart Echo interval when a state change occurs. The reasoning is that a change of state somehow indicates a activity on the peer and that that defines the last know state of the peer. The next Echo will then validate that state (or change it).
> 
> The real problem that this PR tries to fix is that pending requests and context will impact the state of the peer. And that is IMHO the real problem, activity triggered by us should never change the peers state, only responses and timeouts from the peer should impact the peer state.

Yes indeed, the intention of this is not to change the existing logic, which is at the moment mixing echo probe results and contexts to determine the state, and as you mentioned could (and probably should) be fundamentally reconsidered. It only intends to fix a particular situation that can render broken path detection completely dysfunctional in a specific scenario. I'm also not quite sure if it changes the path state logic, but only the scheduling of the echo probes to make sure that the interval changes at state transitions don't result in echoes not being sent.

mgumz:
@fholzhauser is this PR address https://github.com/travelping/smc/issues/37 ?

RoadRunnr:
> @fholzhauser is this PR address [travelping/smc#37](https://github.com/travelping/smc/issues/37) ?

It might, but it is the wrong approach (IMHO).

The underlying problem is that outgoing GTP requests are considered as a livesign of the peer and therefore reset the Echo timer. The correct approach would be to only use GTP packets from the peer (requests and responses) as indication of liveliness.

This PR does not change that. Instead it decouples the GTP echo timer from the other activity indications.

RoadRunnr:
for the 3.x and master branch this is resolved in #428

Is this PR till needed or can we close it?

vkatsuba:
PR https://github.com/travelping/ergw/pull/428 is related to master(`3.X`) version. Current PR is related to `2.X`. Depending on how critical the problem for `2.X` - we can close this PR or provide release `2.8.15` with including fix. However I not sure that this is should be done. I will close current PR and  if this will be necessary to fix it for `2.8.X` too, we always can back to current implementation.

PR #422, opened 2021-07-30
Only put valid pools in session

fholzhauser:
If an upf node pool is configured with only 1 IP version support, the value `undefined` was put into the session for the unsupported pool, causing problems when encoding AAA requests.


RoadRunnr (commented):
This feels like a helper/wrapper would be a good idea

vkatsuba (approved):

PR #432, opened 2021-09-14
Fix gtp_gtp monitors

mgumz:
The branch fix/gtp_path_monitors contains experimental changes to address memory consumption.

This PR is mainly to have a discussion about the proposed changes

### DIFF
--- a/src/gtp_path.erl
+++ b/src/gtp_path.erl
@@ -232,7 +232,7 @@ init([#socket{name = SocketName} = Socket, Version, RemoteIP, Args]) ->
 
 handle_event(enter, #state{peer = Old}, #state{peer = Peer}, Data)
   when Old /= Peer ->
-    ?LOG(debug, "enter state when Old/=New peer, data: ~p", [Data]),
+    ?LOG(debug, "enter state when Old ~p /= Peer ~p", [Old, Peer]),
     peer_state_change(Old, Peer, Data),
     OldState = peer_state(Old),
     NewState = peer_state(Peer),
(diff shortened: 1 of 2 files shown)


RoadRunnr (changes requested):
the actual fix looks ok, but there is far too much noise with all the ?LOG stuff in here.

I would like to have the changes reduced to the fix only.

javiermtorres:
I'm afraid it is the logging that enabled us to see the issue. I really miss a lot of debug logs, specially in the handle_* functions for each gen_server instance. We can work out the current format, which I agree is really hasty and not too pretty, but the logs must remain if we want to solve issues like these in the future. We can hold a meeting during the week to work out the details.

So far we haven't seen any degradation in memory or speed due to the logs, so that's another reason to keep them in. But as I said, we can discuss the specific format. My plan is to add finer grained tracing later on, so specific IPs or IMSIs can be traced with minimal impact on production nodes (although this would still have to be checked by QA). Currently we're enabling specific modules for logging, but we should be able to further filter by e.g. function according to https://erlang.org/doc/apps/kernel/logger_chapter.html#filters

RoadRunnr:
Logging on a full loaded PGW is useless, not only drowns the relevant log entries in the noise of the not relevant messages, it also overloads the system when activated.
Logging on a test branch to catch a specific can be build on per issue case.

if you need gen_server/gen_statem event logging, you can always use https://github.com/travelping/ergw/commit/50ed2c23f4ef79b74da06923ce874f946d998897

javiermtorres:
Logger can have a very fine control about what traces are produced. I don't want to add a new layer on top of something that already works. We can discuss the format, but the traces on handle_X should remain.

mgumz:
@RoadRunnr @javiermtorres: if the only thing that hangs around in the air is log-messages / log-lines: the PR here is against the 2.8er branch and not really relevant for further maybe-it-has-impact situations. so, for now it would be fine with me to keep the log-lines on debug-level in.

whats more crucial: the fix for the memory-situation caused by the monitors. so, either way: please merge against 2.8er branch so i can have a 2.8.15 and thus continue in some of our other products.

PR #433, opened 2021-09-14
rework context teardown on inactivity to include a proper cause

RoadRunnr:
Rework the context inactivity handling to generate a delete GTP
message with a related cause (PDP address inactivity timer expire /
PDN connection inactivity timer expire).

### DIFF
--- a/apps/ergw_core/src/gtp_context.erl
+++ b/apps/ergw_core/src/gtp_context.erl
@@ -418,9 +418,9 @@ handle_event({call, From},
 handle_event({call, From},
 	     {sx, #pfcp{type = session_report_request,
 			ie = #{report_type := #report_type{upir = 1}}}},
-	     State, #{pfcp := PCtx} = Data0) ->
-    Data = close_context(both, up_inactivity_timeout, State, Data0),
-    {next_state, shutdown, Data, [{reply, From, {ok, PCtx}}]};
+	     State, #{pfcp := PCtx} = Data) ->
+    gen_statem:reply(From, {ok, PCtx}),
+    delete_context(undefined, up_inactivity_timeout, State, Data);
 
 %% Usage Report
 handle_event({call, From},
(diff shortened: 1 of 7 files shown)


apps/ergw_core/src/gtp_context.erl
```diff
@@ -418,9 +418,9 @@ handle_event({call, From},
 handle_event({call, From},
 	     {sx, #pfcp{type = session_report_request,
 			ie = #{report_type := #report_type{upir = 1}}}},
-	     State, #{pfcp := PCtx} = Data0) ->
-    Data = close_context(both, up_inactivity_timeout, State, Data0),
-    {next_state, shutdown, Data, [{reply, From, {ok, PCtx}}]};
+	     State, #{pfcp := PCtx} = Data) ->
+    gen_statem:reply(From, {ok, PCtx}),
```
javiermtorres, line 422: Is there a reason not to return a {reply, ...} value instead of manually replying? Is delete_context already returning something?

If this is the case, I'd prefer result values to calls/casts to be explicitly stated in the body of the appropriate handle_X function. It becomes difficult to follow the messages otherwise.

javiermtorres (approved):
Please check the comment in gtp_context.erl. It is just a suggestion, since it'd impact lots of files, but please consider it for new code. Thanks!

apps/ergw_core/src/gtp_context.erl
RoadRunnr: 1. gen_statem does not have a `{reply, ...}` return value, it would need a reply action instead
2. `delete_context/4` is returning `{next_state,...}` already. Passing `{ok, PCtx}` into that would need changes to delete_context and I want to keep the changes as local and minimal as possible.

javiermtorres: 1. Thanks, I was forgetting that it's included in the list of actions.
2. Yes, that is the behaviour I don't like :) I'd like to see a explicit action description in the body of the handle_ function. But I agree that this may not be the best case to start working on this.

PR #434, opened 2021-09-15
redo of #432

RoadRunnr:
redo of #432

* split the actual fix and debug logging into seperate commits
* only collect debug metadata when needed


vkatsuba (approved):
LGTM. If you can, please align code blocks.

PR #439, opened 2021-09-28
move to OTP-23.3

RoadRunnr:


apps/ergw_core/src/gtp_context.erl
```diff
@@ -324,12 +324,13 @@ port_message(Server, Request, Msg, Resent) ->
 
 callback_mode() -> [handle_event_function, state_enter].
 
-init([Socket, Info, Version, Interface,
-      #{node_selection := NodeSelect,
-	aaa := AAAOpts} = Opts]) ->
+init({[Socket, Info, Version, Interface,
```
javiermtorres, line 327: Maybe rename "Interface" to "InterfaceM" and add some doc, just to make sure the casual reader realize that this holds a module ref that determines the behaviour of the gtp_context.

apps/ergw_core/src/gtp_context.erl
```diff
@@ -350,7 +351,8 @@ init([Socket, Info, Version, Interface,
       left_tunnel    => LeftTunnel,
       bearer         => Bearer},
 
-    Interface:init(Opts, Data).
+    {ok, State, LoopData} = Interface:init(Opts, Data),
```
javiermtorres, line 354: Adding some doc detailing that the current state is retrieved from somewhere would be really helpful.

apps/ergw_core/src/gtp_context_reg.erl
```diff
@@ -101,14 +101,9 @@ init([]) ->
...
-    case lists:member(Pid, Links) of
-	true ->
-	    State = maps:update_with(Pid, fun(V) -> [From|V] end, [From], State0),
-	    {noreply, State};
-	_ ->
-	    {reply, ok, State0}
-    end.
+    link(Pid),
+    State = maps:update_with(Pid, fun(V) -> [From|V] end, [From], State0),
+    {noreply, State}.
```
javiermtorres, line 106: If noreply is used, the call will be terminated by timeout. Why not return an ok?

apps/ergw_core/src/gtp_context_reg.erl
```diff
@@ -117,13 +112,18 @@ handle_cast({link, Pid}, State) ->
 handle_cast(_Msg, State) ->
     {noreply, State}.
 
-handle_info({'EXIT', Pid, _Reason}, State0) ->
-    Objs = ets:lookup(?SERVER, Pid),
-    [global_del_key(Key, {Handler, Pid}) || {_, {Handler, Key}} <- Objs],
-    ets_delete_objects(?SERVER, Objs ++ mk_reg_pids(Objs)),
+handle_info({'EXIT', Pid, _Reason}, State) ->
+    Notify = maps:get(Pid, State, []),
+    proc_lib:spawn(
```
javiermtorres, line 117: I still think proc_lib is overkill in this case, but it's not a big deal.
The fun can be finished with any value, there's no need to return an ok since nobody gets it.

javiermtorres (approved):

PR #447, opened 2021-10-21
Bump 3.1 to master

vkatsuba:


vkatsuba:
> Once you "rebase and merge" this PR, the reference will change anyhow. What you want would only work with a merge commit.

Then will bee need rebase stable/3.1 into master, this will save references and push force into master. Then mo make sense in current PR then...

PR #454, opened 2021-11-02
fix the config translation for GTP path options

RoadRunnr:


apps/ergw/src/ergw_compat.erl
```diff
@@ -893,12 +893,23 @@ gtp_proxy_ds_translate_option(Opt, Value) ->
...
+      icmp_error_handling := ICMP,
+      idle_echo := IdleEcho, idle_timeout := IdleTimeout,
       down_echo := DownEcho, down_timeout := DownTimeout} =
-	C0 = translate_options(fun gtp_path_translate_option/2, Values, ?PathDefaults, map),
-    C1 = maps:without([idle_echo, idle_timeout, down_echo, down_timeout], C0),
-    C1#{idle => #{echo => IdleEcho, timeout => IdleTimeout},
-	down => #{echo => DownEcho, timeout => DownTimeout}}.
+	translate_options(fun gtp_path_translate_option/2, Values, ?PathDefaults, map),
+
+    Evs0 = #{icmp_error => warning, echo_timeout => critical},
```
javiermtorres, line 903: Isn't QA using a value of info for the icmp_error?

RoadRunnr: This is used only for the conversion of old (v2.8.x) configs. The old config does not have a value for this and this mirrors the defaults use in ergw_config.

vkatsuba (approved):

PR #456, opened 2021-11-25
fix service normalization in NAPTR records from DNS

RoadRunnr:
The normalize service field is supposed to be ordset. The
normalization instead produced an unsort list. The datatype
itself is the same (a list), but the set operations will not
work as expected on the unsorted list.


vkatsuba (dismissed):

javiermtorres (approved):

vkatsuba (approved):

javiermtorres (commented):
Tests failing?

PR #461, opened 2022-03-14
fix return values of ergw_node_selection:lookup/{2,3}

RoadRunnr:
The two functions have very different return values, but used
an internal helper that would return a value that is unsuitable
for lookup/2.

The helper actually doesn't add much value to the implementation.
Unroll it into the actual functions and fix the return values.

This PR can be backported to all stable branches and should apply without much problems.

### DIFF
--- a/apps/ergw_core/src/ergw_node_selection.erl
+++ b/apps/ergw_core/src/ergw_node_selection.erl
@@ -355,25 +355,27 @@ match(Msg, OrdSPSet) when is_list(Msg) ->
 match(_, _) ->
     [].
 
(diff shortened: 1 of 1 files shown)


javiermtorres:
Is it possible to add a unit test the check that this value is received in ergw_sx_node appropriately?

mgumz:
what kind of effect does this commit fix? an observed issue in production?

at the moment its hard for me to relate this commit to anything in the past which would require a backport.

mgumz:
repository wise: ergw_internal?

RoadRunnr:
wrong repo

